PR #17, opened 2012-09-09
sublime-* files syntax improvements

zuthan:
allow backslash keybinding in sublime-keymap
add comment support for several sublime config file formats


FichteFoll:
All the syntax definitions will likely be remade in some time after `.sublime-syntax` is available on the beta channel.

I think I also covered a few of these issues already but whatever.

Issue #37, opened 2014-01-12
Split the conversion thing from PackageDev?


FichteFoll:
So, today as I was thinking about how I could get back at making this ST3-compatible I kinda questioned if this part of the plugin actually belongs to here. Here is what I ended up with:
- AAAPackageDev is for modifying existing or templating configuration files for ST (syntax definitions, snippets ...)
- There are several other plugins that ease the creation of configuration files for ST so they kinda clash with PackageDev itself (example: SaneSnippets)
- PackageDev does not actually provide easy ways to edit all kinds of settings. Even though the "Convert to ..." build system supports any data file type, including tmPreferences, tmThemes ..., it officially only supports tmLanguage files because it's the only one with syntax highlighting.
- PackageDev itself is a pretty huge package with many languages to handle
- I think it would be easier to maintain the split packages (for me) because I would have direct access to the file conversion stuff

Thoughts?

FYI, I will try to integrate the complete package into ST2 and 3 as my next task.
Edit: ST2 and ST3 compatability is done. Just need to test a few things and create reasonable commits.

DanielRosenwasser:
@FichteFoll 

> * Provide a standalone general-purpose YAML to JSON to Plist to YAML conversion, optionally with a sane YAML highlighting because as you stated the default YAML highlighter sucks
* ...
* Possibly add support for tmPreferences (and tmTheme) files as well

When you say "standalone" would it be possible to extract the functionality to a standalone pip package? I think there would be value in being able to automate the process in a command line setting.

FichteFoll:
Definitely possible, but it wasn't actually something I had in mind. If there is demand for it, I'll see what I can do. Some things need to be refactored though because I pass view objects around a lot and those obivously don't exist if I was to make it stand-alone.

It's still not a priority though, because ... well, it works.

FichteFoll:
This is *very* unlikely to happen, so I'm going to close.

PR #40, opened 2014-01-19
Add tmLanguage for tmTheme files

Adarma:
I have added a .tmLanguage file to highlight .YAML-tmTheme files.


FichteFoll:
@Adarma, as I'm kind of "cleaning up a bit" I suggest you to check out my package [CSScheme](https://packagecontrol.io/packages/CSScheme) which handles this better than yaml ever could, imo.

Would you still need this syntax?

Adarma:
I guess that's good for creating tmTheme files so no need to pull this.

I was also using this package for creating tmLanguage files from YAML. Are you keeping that?

FichteFoll:
You mean aaapackagedev? Yes, that will stay. Or maybe it will get outsourced to a second package,  I don't know yet. See #37.

I have also wanted to use another package for syntax definitions for a while that defines a custom syntax,  supports pattern replacements (for complex regular expressions) and has less overhead. It's called syntaxhighlighttools iirc.

PR #69, opened 2016-01-22
make keys case-insensitive (keyword.other.name)

MattDMo:
Not much to say - just added `(?i)` at the beginning of the regex to make it case-insensitive, as some plugins use camelCase in their key names.

### DIFF
--- a/Syntax Definitions/Sublime Settings.JSON-tmLanguage
+++ b/Syntax Definitions/Sublime Settings.JSON-tmLanguage
@@ -9,7 +9,7 @@
       "begin": "/\\*",
       "end": "\\*/"
     },
-    {  "match": "\"([a-z0-9_.*-]+)\"\\s*?:",
+    {  "match": "(?i)\"([a-z0-9_.*-]+)\"\\s*?:",
        "captures": {
            "1": { "name": "keyword.other.name.sublime-settings" }
        }
--- a/Syntax Definitions/Sublime Settings.tmLanguage
+++ b/Syntax Definitions/Sublime Settings.tmLanguage
@@ -34,7 +34,7 @@
 				</dict>
 			</dict>
 			<key>match</key>
-			<string>"([a-z0-9_.*-]+)"\s*?:</string>
+			<string>"(?i)([a-z0-9_.*-]+)"\s*?:</string>
 		</dict>
 		<dict>
 			<key>include</key>


FichteFoll:
I don't think we should be making any assumptions here anyway since, to my knowledge, all keys are valid, but thanks for the short-hand fix.

Actually, I have a few hours free so I can spend some time on this package.

MattDMo:
Officially, from the JSON spec [ECMA-404](http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf) (page 9), strings are defined as follows:

> A string is a sequence of Unicode code points wrapped with quotation marks (U+0022). All characters may be placed within the quotation marks except for the characters that must be escaped: quotation mark (U+0022), reverse solidus (U+005C), and the control characters U+0000 to U+001F. There are two-character escape sequence representations of some characters. 

> `\"` represents the quotation mark character (U+0022). 

> `\\` represents the reverse solidus character (U+005C). 

> `\/` represents the solidus character (U+002F).  

> `\b` represents the backspace character (U+0008). 

> `\f` represents the form feed character (U+000C). 

> `\n` represents the line feed character (U+000A). 

> `\r` represents the carriage return character (U+000D). 

> `\t` represents the character tabulation character (U+0009). 

> So, for example, a string containing only a single reverse solidus character may be represented as "`\\`".
 
> Any code point may be represented as a hexadecimal number. The meaning of such a number is determined by ISO/IEC 10646. If the code point is in the Basic Multilingual Plane (U+0000 through U+FFFF), then it may be represented as a six-character sequence: a reverse solidus, followed by the lowercase letter u, followed by four hexadecimal digits that encode the code point. Hexadecimal digits can be digits (U+0030 through U+0039) or the hexadecimal letters `A` through `F` in  uppercase (U+0041 through U+0046) or lowercase (U+0061 through U+0066). So, for example, a string containing only a single reverse solidus character may be represented as "`\u005C`". 

> The following four cases all produce the same result: 

> "`\u002F`" 

> "`\u002f`" 

> "`\/`" 

> "`/`" 

> To escape a code point that is not in the Basic Multilingual Plane, the character is represented as a twelve-character sequence, encoding the UTF-16 surrogate pair. So for example, a string containing only the G clef character (U+1D11E) may be represented as "`\uD834\uDD1E`".

I didn't fix the regex up *that* much, but feel free if you have the time :wink:

FichteFoll:
Thanks for the hint.

I'll be using `.sublime-syntax` anyway, so I can utilize variables.

PR #73, opened 2016-03-05
Syntax for Mouse Map 

r-stein:
This PR contains a syntax definition for mouse bindings. It is based on the key binding syntax definition and highlights known keys different than unknown keys.

Example:

pd_mouse_binding


FichteFoll:
Thank you for the contribution.

Please remove the `Support/Sublime Mouse Map.sublime-settings` file. It is not necessary because of the syntax's `fileTypes` list and could cause problems for users trying to set a different syntax for the extensions.

r-stein:
Ok, I have deleted that commit.

PR #75, opened 2016-06-29
Constant.numeric.boolean -> constant.language, added null

MattDMo:
`true`, `false`, and `null` (or their equivalents) are scoped in every language I've worked with as `constant.language`. I wanted to make that consistent here, because (1) syntax highlighting will now be consistent with JSON and other languages, and (2) I wanted to add `null`, which (to my knowledge) is not a Boolean value.


FichteFoll:
That is true.

I still need to revise all of these syntaxes though ...

Issue #71, opened 2016-02-09
feature request: command to auto generate syntax tests based on code in syntax test file


jcberquist:
I am just throwing this out there, as I am not sure about the amount of work it would involve. (I figured you would have a better idea about how much work it would be.) In the abstract at least I think it would very helpful to have a command that could walk over a line of code in a syntax test file and generate syntax tests for that code based on the scope names it currently has. This would be a time saver when tweaking syntax scopes in a `.sublime-syntax` file. So for example you would go from before:

```
<!-- SYNTAX TEST "HTML.sublime-syntax" -->
<html>
```

to after:

```
<!-- SYNTAX TEST "HTML.sublime-syntax" -->
<html>
<!-- <- text.html.basic meta.tag.structure.any.html punctuation.definition.tag.begin.html -->
 <!-- <- entity.name.tag.structure.any.html -punctuation.definition.tag.begin.html -->
<!-- ^ punctuation.definition.tag.end.html -entity.name.tag.structure.any.html -->
```

or some form of the above. I figure it would add a test wherever the set of scope names changed on the line. What do you think?

FichteFoll:
Thanks for the suggestion. That definitely sounds doable, assuming we don't do too much guessing or configuration, which would complicate things.

I'd have to think about how much specific the scope selectors need to be when the scope name changes. Maybe you can think of some more examples or rules?

jcberquist:
Thanks for considering this! I have a bunch of syntax tests that don't test for scope endings/subtractions, which bit me recently; I need to go back and update them and a command like this would save me a lot of grunt work time.

I don't know that I have some rules to share, but my naive thought is that I would start each line with a list of all scopes present there as a baseline. I don't think I would be inclined to look outside the line, as a scope subtraction can easily be added manually if the test needs to ensure that some particular scope is not present, and that seems like it could get really complicated. (This line could be removed if it is not needed, but it would indicate which scopes were present at the start of the line.) Then I would definitely want the full scope names of any scopes added, but perhaps only enough of scope subtractions to ensure their uniqueness. So in the example I gave, I might shorten the subtractions to `-punctuation` and `-entity` respectively, as that is enough to ensure that those particular scopes have ended. I think doing it that way might improve readability. 

In any case, though, my view is that what a command like this would do is provide a good baseline from which to work. It is easy enough to remove test lines or tweak them after they are there, and this would let users avoid the grunt work of tracking down and entering the scope names and all the places on the line the scope set changes in the first place. 

[...]

PR #82, opened 2016-12-17
added keybindings for syntax test development

keith-hall:
added keybindings and a visual helper for syntax test development as referenced to from https://github.com/SublimeText/PackageDev/issues/71#issuecomment-267342418


jcberquist:
@keith-hall: This is minor, but would you consider adjusting the highlight listener to not erase the `current_syntax_test` region if it is unchanged? Currently, as the cursor moves on the test line, the highlight region flickers as it is erased and redrawn.

jcberquist:
@keith-hall: Thanks! Am I right that you don't actually need to erase the region explicitly if you call `view.add_regions()` again with the same key? I think it overwrites what is already there?

keith-hall:
great idea @jcberquist - and thanks for the tip! it works perfectly and saved me lots of time ;)

keith-hall:
out of interest, does anyone ever really work with multiple cursors while writing syntax tests? I don't, and so I've only added support for the 1st cursor in this PR so far.

syntax-test_dev.py
```diff
@@ -0,0 +1,271 @@
...
+
+        col_start, col_end = lines[0][1]
+        if highlight_only_cursor:
+            col_start = self.view.rowcol(cursor.begin())[1]
+            col_end = self.view.rowcol(cursor.end())[1]
+        elif col_end == col_start:
+            col_end += 1
+
+        region = sublime.Region(line.begin() + col_start, line.begin() + col_end)
+        self.view.add_regions('current_syntax_test', [region], 'text', '', sublime.DRAW_NO_FILL)
```
FichteFoll: Imo this should be configurable. Both the style and the scope name.
I already have highlighting from https://packagecontrol.io/packages/SublimeLinter-contrib-sublime-syntax on test lines, so they *might* conflict (although SL's coloring is almost always red).

syntax-test_dev.py
```diff
@@ -0,0 +1,271 @@
...
+    if len(lines) == 0:
+        return (None, line[2])
+    else:
+        return (lines, line[2])
+
+
+class SyntaxTestEventListener(sublime_plugin.EventListener):
+    """Event Listener for responding to keybinding contexts"""
+
+    def on_query_context(self, view, key, operator, operand, match_all):
```
FichteFoll: Is `on_query_context` not active in ViewEventListeners?

syntax-test_dev.py
```diff
@@ -0,0 +1,271 @@
+import sublime
+import sublime_plugin
+from os import path
+import re
+
+
+def get_syntax_test_tokens(view):
```
FichteFoll: Since all "helper" functions revolve around a given view,
it could make sense architecture-wise to group the functions within a class and store instances of these in a way that the plugin interface points (commands and event listeners) can access it.

Just a thought. I don't know if this would end up being better.

FichteFoll:
>out of interest, does anyone ever really work with multiple cursors while writing syntax tests?

I do occasionally, but mostly for aligning which doesn't conflict with this feature.

syntax-test_dev.py
keith-hall: I'm sure when I first tried it, it wasn't... but seems to work now...

syntax-test_dev.py
```diff
@@ -0,0 +1,271 @@
...
+    def run(self, edit, **args):
+        """Available parameters:
+        edit (sublime.Edit)
+            The edit parameter from TextCommand.
+        character (str) = '^'
+            The character to insert when suggesting where the test assertions should go.
+        """
+
+        view = self.view
+        char = args.get('character', '^')
```
keith-hall: I did this for consistency when typing the `-` after `<`, i.e. to test the scope at the comment marker

syntax-test_dev.py
```diff
@@ -0,0 +1,271 @@
...
+            line = get_details_of_test_assertion_line(view, view.sel()[0].begin())
+            if line[0] is None:
+                return False
+            else:
+                return is_syntax_test_line(view, line[2].begin() - 1, True)
+
+        def file_contains_syntax_tests():
+            return is_syntax_test_file(view)
+
+        if key not in locals():
```
keith-hall: thanks for the info - I don't seem to get notifications about comments on my gists.

FichteFoll: Yeah, unfortunately you don't. And you can't even mention people iirc.

keith-hall:
I've made some improvements as suggested, thanks :) I will probably only have time to make more after the weekend, but if I can tomorrow, I will try :)

FichteFoll:
Just a note that I'll most likely be focussing my time on writing the Package Control channel review tool before working on PackageDev.

syntax-test_dev.py
```diff
@@ -44,6 +41,10 @@ def get_details_of_test_assertion_line(view, pos):
     - the assertion characters (2nd item in tuple aka `assertion_colrange`)
     """
 
+    AssertionLineDetails = namedtuple(
```
FichteFoll: I preferred this in the module root (since its instances are being used outside of the function).

PackageDev.sublime-settings
```diff
@@ -0,0 +1,12 @@
+{
+    // what add_region styles to use for the highlighting of the line/positions being tested
+    // valid values are:
+    // "DRAW_EMPTY", "HIDE_ON_MINIMAP", "DRAW_EMPTY_AS_OVERWRITE", "DRAW_NO_FILL",
+    // "DRAW_NO_OUTLINE", "DRAW_SOLID_UNDERLINE", "DRAW_STIPPLED_UNDERLINE",
+    // "DRAW_SQUIGGLY_UNDERLINE", "HIDDEN", "PERSISTENT"
+    "styles_for_syntax_test_highlight": ["DRAW_NO_FILL"],
```
FichteFoll: I prefer `syntax_test_highlight_styles`. Same for `scope`.

syntax-test_dev.py
```diff
@@ -236,11 +237,30 @@ def on_selection_modified_async(self):
...
+        # # unfortunately, the `sublime` module doesn't encapsulate them for easy reference
+        # for style in styles:
+        #     if style in [
+        #         'DRAW_EMPTY', 'HIDE_ON_MINIMAP', 'DRAW_EMPTY_AS_OVERWRITE', 'DRAW_NO_FILL',
+        #         'DRAW_NO_OUTLINE', 'DRAW_SOLID_UNDERLINE', 'DRAW_STIPPLED_UNDERLINE',
+        #         'DRAW_SQUIGGLY_UNDERLINE', 'HIDDEN', 'PERSISTENT'
+        #     ]:
+        #         style_flags |= getattr(sublime, style)
+        for style in styles:
+            # check the style matches the casing and naming used for constants in `sublime.py`
```
FichteFoll: I prefer the explicit check here since the module contains other constants that would not make sense in this context. I know that you hardly ever encounter a situation where a user inserts this in their settings, but I like to be on the safe side when evaluating user input, always.

keith-hall:
I believe I have addressed all the points you raised in your review, @FichteFoll.

I've also got it to suggest partial scopes:
ezgif com-8d856ff244

and cope with comment end markers:
ezgif com-122ca4c117

And proved it works nicely with line comments that also scope the trailing newline:
ezgif com-710925ebe9

FichteFoll:
Looks awesome. Thanks for addressing my feedback so quickly. I intend to do more work with this package before the end of the year (holidays, yay), and will hopefully merge this and work on other parts that have been bothering me for a long while.

PR #87, opened 2017-02-20
Fix: Add MAC OS 'option' key to Sublime Key Map syntax definition.

deathaxe:
I am not an OSX user, but ST complained about 'option' key to be invalid. I guess it is not intented to be?


FichteFoll:
Thanks for the PR, but I'm likely not going to accept it because of the pending rewrite.

deathaxe:
Rewrite using new sublime-syntax? Better solution anyway. ;-)

Issue #67, opened 2015-12-08
Redo ALL the syntax definitions


FichteFoll:
They've needed a new version in a long time, so once `.sublime-syntax` gets out of dev, it sounds like a good idea to start working on this.

Because I do not intend to also include the old files, this would mark a backwards incompatible release and the current version would be split off into a branch/tag. (This has already happened.)

---

Current progress:

- [x] Sublime Text Build System (#125)
- [x] Sublime Text Commands (#124)
- [x] Sublime Text Completions (#123)
- [x] Sublime Text Keymap (#90)
- [x] Sublime Text Macro (#127)
- [x] Sublime Text Menu (new) (#126)
- [x] Sublime Text Mousemap (#122)
- [x] Sublime Text Project (new) (2c3b3d9c70d304cda7339f000099cb42122c9f95)
- [x] Sublime Text Settings (#113)
- [x] Sublime Text Snippet (#88)
- [x] Sublime Text Syntax Definition (new) (be29b0732965b8b9549c88f7f72ec28e3e209879)
- [x] TextMate Preferences Textmate (new) (#88)
- [ ] ~~TextMate Syntax Definition (JSON)~~ (this is legacy stuff and likely not happening)
- [ ] ~~TextMate Syntax Definition (YAML)~~ (this is legacy stuff and likely not happening)

Sub-syntaxes:

- [x] Scope selector (#88)
- [x] Oniguruma Regualar Expression (be29b0732965b8b9549c88f7f72ec28e3e209879)
- [x] Raw Snippets (#88)
- [x] Escaped Regular Expression (Boost) (f3784074519d1aa8f1ce863a6edc650c097f9d04)

evandrocoan:
There is also this problem on `sublime-keymap`:
```
    { "keys": ["A"], "command": "open_auto_completion" }
```

The upper case key `"A"` alone is a valid case and I using it normally, but it is marked as invalid:

![image](https://cloud.githubusercontent.com/assets/5332158/26595716/bb207886-4542-11e7-978b-39fbefba51d4.png)

FichteFoll:
Interesting. They certainly aren't valid with modifiers but it seems like for single characters anything goes (if this does indeed work, which I can't verify right now). 

Please open a new issue for things like this next time, though.

FichteFoll:
With f3784074519d1aa8f1ce863a6edc650c097f9d04, the last remaining non-legacy syntax definition has been rewritten or added.

[...]

PR #88, opened 2017-02-23
Add syntax definitions for Property Lists and Scope Selectors, and Snippets, and update Snippet (Raw) definitions

keith-hall:
making headway against https://github.com/SublimeText/PackageDev/issues/67


FichteFoll:
My intention was to have all non-python related stuff in a `Package` subfolder, with each syntax (and their preferences and syntax test files) in their own directory. This should generally be a better structure than what we have currently, where things are kind of all over the place.

Regarding "regex replacements", I suppose that the name is just misleading since the syntax covers all snippet-like substitutions, which is the correct thing to do. I don't know if I ever had a proper name for these things. Will need to look through the undocs, although I seem to remember referring to this as "snippet replacements" most of the time.

I'm not happy with the "Property List" syntax though, because it includes sublime text (or textmate) specific matches for shellVariables and symbolTransformation (and scope selectors iirc). The proper name would probably be "TextMate Preferences". I also believe that a *general* plist definition should probably be added to the default packages, if that is accepted.

I'll need to take a closer look at this later, but thanks for the PR. 🎉

PR #91, opened 2017-02-27
set syntax when a syntax test file is loaded

keith-hall:
useful for when the syntax def doesn't have any file extensions associated with it, or when the test can't use the proper file extension


FichteFoll:
It might be a good idea to start with returning a namedtuple now. That also has implicit documentation attached. ;)

PR #97, opened 2017-04-03
Command completions

r-stein:
This includes
- completion for command names in keymap and python (run_command("...")) files
- completions of the arguments with auto-detection of the argument content (trigger: `args`)
- a file with meta data about built-in commands

TODO:
- [x] complete the meta data ~~json~~yaml-file

demo:

pd_command_completions


keith-hall:
it took me ages to work out what [this regex](https://github.com/SublimeText/PackageDev/pull/97/commits/eb77ca5c0f2280ecc0ce38ceb11a3236c04f53d1#diff-b11192005bc5eb2a378a96afbabb999fR101) was doing until I realized it was `run_command` backwards ;)

r-stein:
I will add a comment to that. It's a trick we regulary use with LaTeXTools to match the content before the caret with `re.match`.

deathaxe:
Sublime Text holds all available commands in global lists exported by sublime_plugin.py

I am wondering, if these lists can be used to implement the auto-completion feature without the need of a json file with API declarations.

I wrote a little example how to access the required information.

Docstrings would need to be defined by plugin authors in that case.

```python
import re
import inspect


def snake_case(name):
    string = re.sub('(.)([A-Z][a-z]+)', r'\1_\2', name)
    return re.sub('([a-z0-9])([A-Z])', r'\1_\2', string).lower()


def print_signatures(sublime_command_list):
    """Print a list of all ST commands and their arguments."""
    for command in sublime_command_list:
        command_name = snake_case(command.__name__[:-7])
        command_args = inspect.getfullargspec(command.run).args
        command_args = {
            arg: arg for arg in command_args if arg not in ('self', 'edit')
        }
        if command_args:
            print('run_command(\'%s\', %s)' % (command_name, command_args))
        else:
            print('run_command(\'%s\')' % command_name)

        if command.run.__doc__:
            print(command.run.__doc__)


def list_st_commands():
    print_signatures(sublime_plugin.application_command_classes)
    print_signatures(sublime_plugin.window_command_classes)
    print_signatures(sublime_plugin.text_command_classes)
```

keith-hall:
The [proposed solution in this PR already uses `sublime_plugin` to get the details of commands that are implemented in Python](https://github.com/SublimeText/PackageDev/pull/97/files#diff-b520cda9729094b882ca35312835c73dR96) - the JSON / YAML is there for the commands that are built in to ST, which aren't exposed by that module.

deathaxe:
That's just perfect. Sorry didn't see it from my quick look on it.

FichteFoll:
Thanks for your work on this.

Merged it into https://github.com/SublimeText/PackageDev/tree/pulls/97, made some adjustments and will continue to toy around with it when I find the time. I'll definitely include it in the next beta release, however.

FichteFoll:
This has been merged.

PR #101, opened 2017-04-14
don't extend syntax test highlighting past the newline character, plus refactor for efficiency

keith-hall:
this PR ensures that the syntax test highlighting will be consistent with what ST actually asserts against when the test assertions continue past the new line character. (In this circumstance, ST asserts against the scope at the new line character.)

EDIT: i.e. it changes it from this:

before pd

to this:

after pd

### DIFF
--- a/syntax-test_dev.py
+++ b/syntax-test_dev.py
@@ -33,82 +33,169 @@ def get_syntax_test_tokens(view):
         return SyntaxTestHeader(**match.groupdict())
 
 
-def is_syntax_test_file(view):
-    """Determine if the given view is a syntax test file or not."""
+class SyntaxTestHighlighterListener(sublime_plugin.ViewEventListener):
+    header = None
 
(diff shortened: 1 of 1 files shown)


FichteFoll:
FYI, I wrote a utility fuction for `sublime_lib` that does this: [`sublime_lib.view.coorded_region`](https://github.com/SublimeText/PackageDev/blob/cf3c44d19f019382c764a264bb6b3962429373db/sublime_lib/view/_view.py#L164-L190).

keith-hall:
Now includes a refactor to improve efficiency and fix #104.

FichteFoll:
Thanks for the refactor. I like this a lot better, actually.

Fixed an issue where both, start and end column of a syntax test, would extend beyond the line to be tested with 9c4d5a94021912dd87edc41276d9766bfe5be942.

Issue #92, opened 2017-02-28
[feature request] highlight relevant capture group


keith-hall:
when writing syntax definitions, one quite often uses multiple capture groups. It can get quite tricky to find/remember where each group starts and which number it is.
I therefore propose that we add functionality similiar to how the syntax test highlighting helper works, and when the caret is on a `captures: n:` line in a `sublime-syntax` file, that the relevant group be highlighted in the corresponding `match:` line

```yaml
  - match: ((a)?(b)cde(fgh))(ij)
    captures:
      1: where is capture group 1 again?
```

keith-hall:
here is the code I am using atm, far from PR-ready quality, but it works:

```python
def find_scope_within_region(view, region, scope_selector):
    for match in view.find_by_selector(scope_selector):
        if region.contains(match):
            yield match

def find_scope_before_region(view, region, scope_selector):
    prev_match = None
    for match in view.find_by_selector(scope_selector):
        if match.begin() < region.begin():
            prev_match = match
        else:
            break
    return prev_match

class HighlightCaptureGroupsViewEventListener(sublime_plugin.ViewEventListener):
    def on_selection_modified_async(self):
        if self.view.match_selector(self.view.sel()[0].begin(), 'source.yaml.sublime-syntax meta.expect-captures'):
            # get the capture group number
            group_num_region = next(find_scope_within_region(self.view, self.view.line(self.view.sel()[0].begin()), 'constant.numeric'), None)
            if group_num_region:
                group_num = int(self.view.substr(group_num_region))
                # find the region above the captures where a regex is expected
                check_region = find_scope_before_region(self.view, group_num_region, 'meta.expect-regexp')
                # ignore comments
                regex_pattern_regions = find_scope_within_region(self.view, check_region, 'source.regexp.oniguruma - comment')
                for region in regex_pattern_regions:
                    pos = region.begin()
                    while pos < region.end():
                        char = self.view.substr(pos)
                        if char == '\\': # skip escape chars
                            pos += 1
                        elif char == '(':
                            if self.view.substr(pos + 1) != '?': # if it is a capture group
                                group_num -= 1
                                if group_num == 0:
                                    pos_start = pos
                        elif char == ')':
                            if group_num == 0:
                                pos_end = pos + 1
                                region = sublime.Region(pos_start, pos_end)
                                self.view.add_regions('current_capture_group', [region], 'text', '', sublime.DRAW_NO_FILL)
                                return
                            elif group_num < 0:
[...]

PR #102, opened 2017-04-20
Highlight regex capture groups when in the captures: section.

Thom1729:
Implementation for #92.

Features:
- Works with multiple selections (highlighting multiple capture groups).
- Handles nested capture groups and correctly avoids non-capturing groups.
- Should gracefully handle erroneous input, although it doesn't point it out.

Known shortcomings:
- We assume that the regexp associated with a `capture` is the last regexp before it in the file. There is surely a better solution.
- There is a bug: when you highlight multiple `capture` lines in a single selection, only the capture group associated with the first line is highlighted. All relevant capture groups should be highlighted.

Other remarks:
- This highlights the zeroth capture group, which is arguably a bad idea.
- The highlighting uses the `comment` scope, which I specified more or less at random. I'm not sure what the convention is.
- Some of the manual character-twiddling could be avoided with a little more sophistication in the regexp syntax (e.g. using `meta.group.capturing` for capturing groups).

### DIFF
--- a/syntax_def_dev.py
+++ b/syntax_def_dev.py
@@ -509,11 +509,15 @@ class SyntaxDefCompletions(sublime_plugin.EventListener):
 
         return inhibit(self.base_completions + completions)
 
+
 class SyntaxDefRegexCaptureGroupHighlighter(sublime_plugin.ViewEventListener):
     @classmethod
     def is_applicable(cls, settings):
         syntax = settings.get('syntax')
-        return syntax == 'Packages/PackageDev/Package/Sublime Text Syntax Definition/Sublime Text Syntax Definition.sublime-syntax'
+        return syntax == (
+            'Packages/PackageDev/Package/Sublime Text Syntax Definition' +
+            '/Sublime Text Syntax Definition.sublime-syntax'
+        )
 
     def on_selection_modified(self):
         self.view.add_regions('captures', list(self.get_regex_regions()), 'comment')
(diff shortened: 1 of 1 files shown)


syntax_def_dev.py
```diff
@@ -508,3 +508,77 @@ def inhibit(ret):
...
         return inhibit(self.base_completions + completions)
+
+class SyntaxDefRegexCaptureGroupHighlighter(sublime_plugin.ViewEventListener):
+    @classmethod
+    def is_applicable(cls, settings):
+        syntax = settings.get('syntax')
+        return syntax == 'Packages/PackageDev/Package/Sublime Text Syntax Definition/Sublime Text Syntax Definition.sublime-syntax'
+
+    def on_selection_modified(self):
+        self.view.erase_regions('captures')
```
keith-hall: it would be better to only erase the regions when they no longer need to be drawn - i.e. the selection/carets move out of the captures area - to avoid flickering. This is how the syntax test highlighting works.

syntax_def_dev.py
```diff
@@ -508,3 +508,77 @@ def inhibit(ret):
...
+        self.view.add_regions('captures', list(self.get_regex_regions()), 'comment')
```
keith-hall: you could use settings to get the user's preferred scope name and styles - again, just like the syntax test highlighting does.

syntax_def_dev.py
```diff
@@ -508,3 +508,77 @@ def inhibit(ret):
...
+                if p in { '(', '(?' }:
+                    depth += 1
+                else:
+                    if depth == 0:
+                        end = i+1
+                        break
+                    else:
+                        depth -= 1
+
+            if end != None:
```
keith-hall: as Travis says, this should be `is not None`

FichteFoll:
Please address all comments by @keith-hall and the warnings emitted by flake8. Other than that, this looks pretty good.

keith-hall:
Oh yeah, sorry @Thom1729, I forgot to say I like the way you implemented it :)
About the two shortcomings, I personally don't think anyone should ever have the captures above the corresponding regex, as its a bit odd, so I don't see that as a problem. Also the main use I have is with a single selection that doesn't span multiple lines so I don't see it as a problem that it only highlights the capture relating to the first selected line.
Thanks for taking the time to work on this supremely useful feature :)

Thom1729:
Everything should be good except for the settings. That, I think, should be factored out from `HighlightTestViewEventListener`, but I'm not sure where in the project would be appropriate. Somewhere in `/sublime_lib`?

FichteFoll:
See https://github.com/SublimeText/PackageDev/blob/cf3c44d19f019382c764a264bb6b3962429373db/syntax-test_dev.py#L298-L300.

Thom1729:
I committed a copy-paste of the existing implementation. It may be worth a new issue to abstract out some of Sublime's flag enums, including this one.

FichteFoll:
Thanks. This pull request has been integrated manually.

FichteFoll:
Thanks for this; squashed into 3df7d45b74890a9e516568e910eab3ca7d15acb2.

PR #105, opened 2017-05-02
use oniguruma regex syntax inside tmPreferences

keith-hall:
ST uses the oniguruma regex engine/syntax inside `.tmPreferences` files, so it makes sense to change the syntax highlighting to use it, so that we will have correct/more accurate highlighting on things like [`\k<test+0>` where the capture group backreference can optionally target a particular relative recursion level](https://github.com/kkos/oniguruma/blob/36b89b9ba5c3dd90f7791c8d86517a4d8d82434c/doc/RE#L249)

### DIFF
--- a/Package/Sublime Text Syntax Definition/syntax_test_onig-regexp
+++ b/Package/Sublime Text Syntax Definition/syntax_test_onig-regexp
@@ -488,6 +488,6 @@ where escape characters are ignored.\).
 (?-x)
 
 check if all scopes are cleared:
-# <- (meta.literal) | - meta
+# <- meta.literal - meta meta
 
 # <- - meta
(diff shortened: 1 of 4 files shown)


Package/Sublime Text Syntax Definition/syntax_test_onig-regexp
```diff
@@ -471,6 +488,6 @@ where escape characters are ignored.\).
 (?-x)
 
 check if all scopes are cleared:
-# <- meta.literal - meta meta
+# <- (meta.literal) | - meta
```
FichteFoll: Afaik this breaks syntest, but removing the parens should be fine.

keith-hall: I think I didn't mean to change that test, not sure where it came from - I've undone the change now

PR #111, opened 2017-06-01
add indentation rules for Oniguruma RegExp syntax

keith-hall:
this PR adds indentation rules to the Oniguruma RegExp syntax, see the comments in the file for details. It also fixes a bug with `indentParens` (set in the Default package) that meant that typing `- match: (?=\))` <kbd>Enter</kbd> in a `.sublime-syntax` file would unindent the next line.

### DIFF
--- a/Package/Sublime Text Syntax Definition/Oniguruma RegExp Indentation Rules.tmPreferences
+++ b/Package/Sublime Text Syntax Definition/Oniguruma RegExp Indentation Rules.tmPreferences
@@ -1,0 +1,74 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>scope</key>
+	<string>(source.regexp, source.yaml.sublime.syntax meta.expect-regexp) - comment</string>
+	<key>settings</key>
+	<dict>
+		<key>comment</key>
+		<string>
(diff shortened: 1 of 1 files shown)


Package/Sublime Text Syntax Definition/Oniguruma RegExp Indentation Rules.tmPreferences
```diff
@@ -0,0 +1,74 @@
...
+			        |   M-\\C- \d+
+			        )
+			    |   \\.                    # a single escape character
+			    )*+
+			)
+			(?<balanced_paren>
+			    \(                         # an open paren
+			    \g<not_paren>              # followed by anything that's not a group-related paren
+			    \g<balanced_paren>*+       # followed by any number of nested parens
+			    \)                         # a closing paren
```
FichteFoll: Shouldn't non-paren characters be allowed before the closing paren, too?

FichteFoll:
Thanks. Left a comment for consideration.

Also nice usage of possessive quantifiers (and an atomic group). I did this too in the YAML syntax highlighter before sregex was a thing and I had to remove them since they aren't supported.

Package/Sublime Text Syntax Definition/Oniguruma RegExp Indentation Rules.tmPreferences
keith-hall: yes, they should - nice spot, thank you :) pushed a new commit that addresses this 👍

FichteFoll:
Cherry-picked as 6b1aaa8b0af23234948b51ded4e116f5acb8bf79 and acb2f8b008ed4091d1fa693c0fc3007fd6ea7184.

Later revised with 2b59ee2ccc4e6f8c40742e65e291e23498262b3e.

PR #115, opened 2017-07-08
Sublime Text Settings suggestions

deathaxe:
Here are some more suggestions for the settings module

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -116,7 +116,13 @@ def int_or_float(string):
 
 
 def _settings():
-    return sublime.load_settings("PackageDev.sublime-settings")
+    """Return global package settings object or load it first if required."""
+    try:
+        return _settings.settings
+    except AttributeError:
+        l.debug("Loading PackageDev.sublime-settings")
+        _settings.settings = sublime.load_settings("PackageDev.sublime-settings")
+        return _settings.settings
 
 
 class SettingsListener(sublime_plugin.ViewEventListener):
--- a/settings.py
+++ b/settings.py
@@ -51,6 +51,8 @@ VALUE_SCOPE = (
     "meta.expect-value | meta.mapping.value | "
     "punctuation.separator.mapping.pair.json"
 )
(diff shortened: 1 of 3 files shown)


settings.py
```diff
@@ -116,7 +116,13 @@ def int_or_float(string):
 
 
 def _settings():
-    return sublime.load_settings("PackageDev.sublime-settings")
+    """Return global package settings object or load it first if required."""
```
FichteFoll: FYI, ST caches settings in its internal API already, which is why calling `.load_settings` repeatedly is acceptable, but I'm okay with the change. The objects were still being created for no real benefit.

```py
>>> sublime.load_settings("PackageDev.sublime-settings").settings_id == sublime.load_settings("PackageDev.sublime-settings").settings_id
True
```

deathaxe: > ST caches settings in its internal API already

I know. That's why I found creating a new python object on each call somehow a waste. This is what I meant with creating "garbage" as the objects need to be collected by the garbage collector again and again. I recently found this way to cache the object in one of @r-steins implementations and found it a quite interesting construct.

FichteFoll:
Regarding quotation marks, I have the following guidelines:

- Generally, use double quotes. This is prevalently for arbitrary "text", like messages printed or constructed file paths.
- Use single quotes for symbol-like purposes, such as dictionary keys (or setting keys), function parameters (e.g. `open(mode='r')` or `str.decode('utf-8')`). Basically, situations in where the exact content of the string matters.
- Additionally, use single quotes if you'd have to escape double quotes within the string.

I find this guideline to be a good balance act between arbitrarily choosing either as the default and adding semantic meaning to a string based on the quotes used.

I try to follow these for my personal use cases, but I don't necessary go around changing other people's contributed code to adhere to them, which is also why I only changed the quotation marks for strings that I edited anyway in that commit.

FichteFoll:
Cherry-picked all but the quotation commit.

PR #117, opened 2017-07-10
Sublime Settings: Some fixes and a proposal for your latest changes.

deathaxe:
This PR fixes a couple of issues introduced by changes made to the code of PR #113

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -106,15 +106,26 @@ def get_value_region_at(view, point):
     return None
 
 
-def sorted_completions(completions):
-    return list(sorted(completions, key=lambda x: x[0].lower()))
-
-
-def int_or_float(string):
+def decode_value(string):
+    """Decode string to python object with unrestrictive booleans."""
+    if string.lower() == "true":
+        return True
+    if string.lower() == "false":
+        return False
     try:
         return int(string)
     except ValueError:
         return float(string)
+
+
+def format_completion_item(value):
+    """Create a completion item with its type as description."""
+    return ("{0}  \t{1}".format(value, type(value).__name__), value)
+
+
+def sorted_completions(completions):
+    """Sort completions case insensitive."""
+    return list(sorted(completions, key=lambda x: x[0].lower()))
 
 
 def _settings():
@@ -521,7 +532,7 @@ class KnownSettings(object):
                 "{0}  \tsetting".format(key),
                 self._key_snippet(key, value, eol=eol)
             ] for key, value in self.defaults.items()]
-        return (completions, sublime.INHIBIT_WORD_COMPLETIONS)
+        return completions, sublime.INHIBIT_WORD_COMPLETIONS
 
     @staticmethod
     def _key_snippet(key, value, bol="", eol=",\n"):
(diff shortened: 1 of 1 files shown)


settings.py
```diff
@@ -752,7 +752,7 @@ def _color_scheme_completions(view):
                 "{}  \tPackage: {}".format(file_name, package),
                 scheme_path
             )
-            completions.add(item)
+            completions.append(item)
         return completions
```
FichteFoll: `sorted_completions` returns a list and should have been used here. Seems I just forgot to add it.

settings.py
```diff
@@ -782,5 +782,5 @@ def _theme_completions(view):
                 "{}  \ttheme".format(theme),
                 theme
             )
-            completions.add(item)
+            completions.append(item)
```
FichteFoll: Changing this to a list without re-adding the `if item not in completions` check will result in duplicates in the results.
Sets are the proper structure to use here (and imo severely underused in many projects) and a list should be returned at the end of the function (see comment above).

settings.py
```diff
@@ -748,11 +748,8 @@ def _color_scheme_completions(view):
             if any(hide in scheme_path for hide in hidden):
                 continue
             _, package, *_, file_name = scheme_path.split("/")
-            item = (
-                "{}  \tPackage: {}".format(file_name, package),
```
FichteFoll: Themes are specified by the basename alone, because they can be patched by other packages. Color schemes on the other hand are immutable.

I agree that the Package adds some noise and it's probably not necessary, though imo it also adds value when you copied a color scheme to your User package for example. Maybe just removing `Package:` from the hint is enough here?

settings.py
deathaxe: Sorting completions on return of `value_completions()` is just enough. Why sorting twice? You just should not return a set() here which may be accessed by index later in `value_completions()` .

With a second look into `value_completions()` ...
1. I find it a bit messy (even pylint complains about too much branches) and 
2. The whole stuff beginning with line 630 is useless to apply to color scheme / theme completions. It just needs to apply to comments read from comments or the "default" completion.

deathaxe: Just found some other issues within value_completions(). Will have a deeper look into that.

deathaxe: Your changes in value_completions() broke a couple of things. 😠 

- automatic adding/stripping quotes to/from completions is broken,
- completions for float values (e.g. "dpi_scale") are broken
- some completions are sorted several times

somehow much more error prone than before

settings.py
deathaxe: Just found the `if item not in completions` has gone (or was there). I was not yet aware of sets being able to automatically make "lists" unique. Cool feature - again learned something. 😄 

I am on trying to find a common way to create these items no matter of source without too many type conversions and list recreations.

Quotation issue seems to apply to list elements primarily.

FichteFoll:
I'll take a look again over the weekend, probably.

deathaxe:
Here are some suggestions to hopefully fix most of the issues. Please check, whether the back tick quoted completions - especially found lists - still work as expected. I'd suggest to use list items for completions rather than the list itself. This applies to list default values and lists found in comments.

FichteFoll:
FYI, I started merging this a few days ago but am not through with it just yet. I'll cherry-pick any following commits, probably.

FichteFoll:
Thanks. Merged all but the last, which I cherry-picked.

I also more strictly pushed set usage and combined comment and default completions.

PR #119, opened 2017-07-28
Sublime Settings: Fix default completion of empty default string

deathaxe:
If the `default` is an empty string it is not None but has nothing to offer.

This causes a value like "font_face" being completed with `\t  (default) str`

### DIFF
--- a/plugins_/settings/known_settings.py
+++ b/plugins_/settings/known_settings.py
@@ -656,7 +656,7 @@ class KnownSettings(object):
             {(trigger, contents), ...}
                 A set of all completions.
         """
-        if not default:
+        if default is None or default is "":
             return None
         elif isinstance(default, bool):
             return {format_completion_item(True), format_completion_item(False)}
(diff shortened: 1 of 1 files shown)


FichteFoll:
This will also trigger when the default is `0`. Just add an explicit check for the empty string somewhere.

FichteFoll:
Or False, even.

plugins_/settings/known_settings.py
```diff
@@ -34,7 +34,7 @@ def format_completion_item(value, default=False):
                                    type(value).__name__,
                                    default_str),
             # 'cast' dicts to frozen sets, because those are hashable
-            frozenset(value.items()) if isinstance(value, dict) else value)
+            frozenset(value) if isinstance(value, dict) else value)
```
FichteFoll: This is a curious case. Maybe we should discuss what to do with dict settings in general.

deathaxe: If you don't want to try to support nested key's completion, we at most should offer the keys of the first level as completion if any at all. I think it may even not make much sense at all as those structures might differ too much in structure.

We need to ensure not to offer top level settings keys in such a value scope if no value completion is offered.

FichteFoll:
Something went *slightly* wrong when merging this, which resulted in the last two comments actually not being merged. Since I already fixed the bug from the one commit, I just cherry-picked the other.

Anyway, this is merged.

PR #125, opened 2017-08-13
Rewrite "Build System" syntax

r-stein:
This rewrites the "Build System" syntax based on the new keybinding syntax. It also adds completion files.


Package/Sublime Text Build System/Sublime Build System.sublime-syntax
```diff
@@ -0,0 +1,400 @@
...
+                scope: punctuation.separator.placeholder.regex-replace.end.placeholder.sublime-build
+                set:
+                  - match: \}
+                    scope: keyword.other.block.end.placeholder.sublime-build
+                    pop: true
+                  - match: '[gimsx-]*'
+                    scope: constant.other.regex-modifier.placeholder.sublime-build
+                  - match: \S
+                    scope: invalid.illegal.unexpected-token.placeholder.sublime-build
+              - include: scope:source.regexp-replacement
```
r-stein: This end the next line are not json escaped. Is there a similar scope, but json escaped?

FichteFoll: I believe we only have this in the legacy format. It's listed in #67 (escaped boost regex).

FichteFoll: Actually, check what the key bindings are using for context checks.

r-stein: The keybindings uses the escaped regular expressions from this package, but this does not contain the replacements, i.e. only `A` of `${file/A/B/g}`. So I think we also need to rewrite that.

https://github.com/SublimeText/PackageDev/blob/v3.0.0-beta.3/Package/Regular%20Expression%20(Escaped).tmLanguage

PR #135, opened 2017-09-09
suplime-menu snippets

deathaxe:
See commit messages, please.


FichteFoll:
Thanks for the addition. Those already existed for keymaps, but I was getting lazy and didn't add them for any other file.

This is now fixed in a new release with proper completions for *all* JSON resource files and properly documented in the wiki: https://github.com/SublimeText/PackageDev/wiki#snippets-and-completions

deathaxe:
> lazy

You are joking!?

PR #137, opened 2017-09-11
Add some padding to edit settings phantoms

deathaxe:
The padding adds some spacing to the left and right of the edit symbol to make the button look better. The symbol is rendered centered. 0.2rem should add a dpi/font-size independent relative spacing which grows with the text.

### DIFF
--- a/plugins_/settings/__init__.py
+++ b/plugins_/settings/__init__.py
@@ -57,7 +57,7 @@ PHANTOM_TEMPLATE = """
 <style>
     body {{
         margin: 0;
-        padding: 0;
+        padding: 0 0.2rem 0 0.2rem;
     }}
     a {{
         text-decoration: none;
--- a/plugins_/settings/__init__.py
+++ b/plugins_/settings/__init__.py
@@ -56,8 +56,9 @@ PHANTOM_TEMPLATE = """
 <body id="sublime-settings-edit">
 <style>
     body {{
+        background-color: var(--background);
         margin: 0;
-        padding: 0 0.2rem 0 0.2rem;
+        padding: 0;
     }}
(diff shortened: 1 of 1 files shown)


FichteFoll:
Can you provide before and after screenshots and the theme & color scheme you use?

deathaxe:
before

screenshot_before

after

screenshot

Using a modified version of _AYU dark_ color scheme together with _Adaptive Theme_, but this shouldn't matter as padding and borders are controlled by the stylesheet being used. 

Maybe the chosen font and/or OS may make a difference.

FichteFoll:
I see. So, the reason why it doesn't bother me is because my phantoms use the same background color as the color scheme and phantoms have some padding around them by default. With the change, padding would effectively double because this "box" didn't exist for me.

Would changing the phantom's background color to the color scheme's work for you or do you have an argument against this?

deathaxe:
Removing the background would be an appropriate alternative.

deathaxe:
> Edit: popupCss in the specific theme (doesn't provide phantomCss):

You can not assume all color schemes do!

Therefore PackageDev needs to ensure either correct padding is added in case phantoms are rendered with a different background or it needs to ensure no background is rendered for the added phantoms at all.

FichteFoll:
>it needs to ensure no background is rendered for the added phantoms at all

This is what I decided on. In addition to what you did in your last commit, I also target html, because color schemes may have additional padding on the html tag, making its background color visible even if I override it for the body. Since I set the padding to `0` with that block as well, I don't think we have anything to fear anymore.

deathaxe:
Just saw the PR was closed but missed the commit.

Issue #144, opened 2017-09-15
TAB don't work dialog displayed with spaces as indention


deathaxe:
I just tried to play around with https://github.com/sublimehq/Packages/pull/1126 when I found PackageDev complaining about using TABS for indention doesn't work again and again, each time I click into the view.

![screenshot](https://user-images.githubusercontent.com/16542113/30502748-d2d2aa06-9a67-11e7-8d12-b494205916dc.png)

The fact is: The file uses 4 spaces for indention. Even a call to "Convert tabs to spaces" does not help.

y0ssar1an:
With this setting, you can see tabs and spaces:
```js
"draw_white_space": "all"
```
It can help you find any tabs that might've snuck in.

FichteFoll:
The code internally only checks for the "translate_tabs_to_spaces" setting internally, iirc. Can you check what the value of that is in your scenario?

deathaxe:
> With this setting, you can see tabs and spaces:

Everyone knows that. 😉

> Can you check what the value of that is in your scenario?

My global value in Preferences is `true`

Query view setting from console also shows `true`

```
>>> view.settings().get("translate_tabs_to_spaces")
True
```

Seems to be an temporary effect. To check that, I added some debug messages to the `SyntaxTestHighlighterListener`'s methods.

With one view open, which contains the syntax_test_git_tag, I closed and restarted ST. It contains the following output without error message. Everything seems ok.

```
is_applicatable(settings_id : 6) : translate_tabs_to_spaces True
__init__(view_id: 14, settings_id : 6) : translate_tabs_to_spaces True
is_applicatable(settings_id : 6) : translate_tabs_to_spaces True
is_applicatable(settings_id : 6) : translate_tabs_to_spaces True
is_applicatable(settings_id : 6) : translate_tabs_to_spaces True
is_applicatable(settings_id : 6) : translate_tabs_to_spaces True
is_applicatable(settings_id : 6) : translate_tabs_to_spaces True
is_applicatable(settings_id : 6) : translate_tabs_to_spaces True
is_applicatable(settings_id : 6) : translate_tabs_to_spaces True
```

If I now click into the console, the error dialog is displayed and the following lines are added to console.

```
is_applicatable(settings_id : 6) : translate_tabs_to_spaces False
__del__(view_id: 14, settings_id : 6) : translate_tabs_to_spaces False
is_applicatable(settings_id : 6) : translate_tabs_to_spaces False
[...]

PR #145, opened 2017-09-16
Improve SyntaxTestHighlighterListener header handling

deathaxe:
While debugging #144 I found the following possible (small) improvements for the event SyntaxTestHighlighterListener.

1. The `self.header` is an object attribute and should be therefore declared in `__init__()`
2. The `self.header` is queried first in `__del__()` to avoid reading settings, if we are not within a syntax_test_...
3. Save 2 lines of code when checking syntax_test_ file name


FichteFoll:
Unfortunatey an interactive rebase flattened the merge, but this has been merged as a757851d49e4c9af45e6088eeeb9ca1f5d00d241.

PR #148, opened 2017-09-19
Remove some unnecessary captures.

deathaxe:
Captures take some time to be handled and thus should be avoided where possible.

### DIFF
--- a/Package/Sublime JSON/Sublime JSON.sublime-syntax
+++ b/Package/Sublime JSON/Sublime JSON.sublime-syntax
@@ -84,7 +84,7 @@ contexts:
     - include: null-pop
     - include: sequence-pop
     - include: mapping-pop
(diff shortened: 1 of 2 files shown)


keith-hall:
Actually it's a fairly simple optimization to ignore captures if none of the captures in the expression are used - [syntect](https://github.com/trishume/syntect) does this, for example.

deathaxe:
ST seems not. I did several benchmarks with my CNC package. It turned out avoiding captures can easily increase performance by 10 to 15%.

FichteFoll:
I believe it is misleading to use captures without the intend of actually capturing them, regardless of runtime performance, so thanks for that.

FichteFoll:
Merged (and rebased) as e5cc2dd50968d5ac89136602c71d636582389878.

PR #149, opened 2017-09-19
Fix float settings values completion

### LINKED ISSUE #174: Is there a mechanism for embedding DSLs in existing highlighters?
This is much more easily explained with an example, and I'll use the one I care about, but the question's sort of general by its nature

Some languages - recently including Javascript - have a mechanism either explicitly or sort of nudge-nudge-wink-wink generally meant for the creation of DSLs.  In Javascript, this is the `tagged template literal`.

Hi, I'm Troy McClure.  You may remember me from such films as that annoying ticket yesterday, or begging for help with easy things on IRC

For the sake of a short explanation that only I will find funny, let's say that I've come up with some domain specific language for describing pants.

So if you're *fancy*, in Javascript, these days, you're supposed to 

```javascript
const jeans = pants`holes   : yn
                    color   : [blue|black]
                    fit     : [tight|loose|jnco]
                    bottoms : [tight|flare|bell]`;
```

Now, obviously I can't expect Sublime's HTML/JS highlighter to know about or accomodate my awesome, category defining new pants DSL.  After all, we're not even releasing it until after the investor party next week

All the same, Sublime is my one true love, so when I'm spending the rest of my days working in `PantsL`, on those days I have to embed it on the web, I still want it hilit in all its glory.

Is there a way for me to write a highlighter which adds itself to some other highlighter?  Intrudes, as it were?

deathaxe:
This commit fixes #174 with following changes.

1. Match 1. as float even without a decimal digit in order to not pop from numbers stack to early while typing.
2. Extend the meta.mapping.value scope until the next none space character.
3. Extend the VALUE_SCOPE by `invalid.illegal.expected-value` to enable value completion in empty region between `:  ,`.
4. Make sure to exclude VALUE_SCOPE from KEY_COMPLETIONS_SCOPE to avoid key completions within the value region.

### DIFF
--- a/Package/Sublime JSON/Sublime JSON.sublime-syntax
+++ b/Package/Sublime JSON/Sublime JSON.sublime-syntax
@@ -23,7 +23,7 @@ variables:
         | [1-9] \d*
       )
       (?:
-        (?: \. \d+ )?
+        (?: \. \d* )?
         (?: [eE] [+-]? \d+ )?
       )?
     )
@@ -416,7 +416,7 @@ contexts:
   mapping-value-meta:
     - clear_scopes: 1
     - meta_scope: meta.mapping.value.json
-    - match: ''
+    - match: '(?=\S)'
       pop: true
 
 
--- a/plugins_/settings/region_math.py
+++ b/plugins_/settings/region_math.py
@@ -1,13 +1,14 @@
 
 # match top-level keys only
-KEY_SCOPE = "entity.name.other.key.sublime-settings"
-KEY_COMPLETIONS_SCOPE = (
-    "meta.settings-mapping.sublime-settings - comment - meta.setting-value.sublime-settings"
-    " | " + KEY_SCOPE
-)
 VALUE_SCOPE = (
     "meta.expect-value | meta.setting-value.sublime-settings"
-    " | punctuation.separator.mapping.pair.json"
+    "| invalid.illegal.expected-value | punctuation.separator.mapping.pair.json"
+)
+
+KEY_SCOPE = "entity.name.other.key.sublime-settings"
+KEY_COMPLETIONS_SCOPE = (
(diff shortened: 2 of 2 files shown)


FichteFoll:
1) This is good.
2) I'm not sure about this. Expanding the value scope beyond the actual value includes matching comments and whitespace, which aren't actually part of the value. On the other hand, if the value is a mapping or a sequence, whitespace within that gets colored (which makes sense and shouldn't be debated here).
  An alternative would be to add a `meta.expect-comma` scope for the region after the value and the comma, including the latter. I believe that better matches our intent.
3) 👍 
4) I specifically didn't exclude value scope because it breaks key completions in project files, where the entire settings dict is a value itself already. We might have to anchor the value scope to a settings mapping meta scope.

deathaxe:
2. Thought about something like the `meta.expect-comma`, too. Would fit into the system. Not sure where and how to do changes to not break anything when adding something like that. I think the meta scopes tend to become too detailed and small.
4. Damn. It's the same thing as with menus and the `meta.mapping.value` stuff. Maybe its a good idea in general to create language specific `meta.key` and `meta.value` scopes instead of relying on json. This way JSON compatibility would be kept but the more specific needs of certain languages could be addressed.

FichteFoll:
Don't worry, I have ideas about solving 2 and 4. I also wrote the imo well-structured albeit complex Sublime JSON definition, so I know where the I need to insert the new contexts.

Might only get around to doing it tomorrow, though.

deathaxe:
> well-structured albeit complex

It introduces interesting structures I've not seen anywhere else so far. I like the way you base anything onto JSON and reduce the repetition of matches/scopes. But it comes a long with quite a couple of complex structures - that's right.

FichteFoll:
Fixed in f0bc60dba641c76240ae782d2e7fbc6082471dc2.

PR #151, opened 2017-09-20
Fix syntax assignment in syntax_test files

### LINKED ISSUE #150: Syntax in SYNTAX TEST files is not applied
PackageDev seems to actively apply the syntax to a _syntax_test_ ..._ file, which works fine if the header looks like that:

```
; SYNTAX TEST "Packages/CNC Sinumerik 840D language support/gcode/s840d_gcode.sublime-syntax"
```

If the header contains the syntax file only,  _Plain Text_ is assigned as syntax.

```
; SYNTAX TEST "s840d_gcode.sublime-syntax"
```

If PackageDev is disabled, in both cases Sublime Text assigns the correct syntax to the opened file.

Therefore I'd expect the same behavior with PackageDev enabled.

deathaxe:
A SYNTAX TEST header can contain a relative path or basename of a syntax file. The later one was not respected before this commit.
This commit modifies the AssignSyntaxTestSyntaxListener in following ways to fix that issue:

1. Plain Text is used as safe default when reading a view's syntax.
2. If the view syntax differs from the SYNTAX TEST header, a sublime-syntax file, matching the basename of the header is searched.
   a) If the list is empty, Plaint Text scope is applied.
   b) If none of the found syntaxes matches the current view's one, the first found syntax is applied.

_Note:_

Even if the HEADER defines a relative path, which does not match the view's syntax file, the best matching candidate is applied without respect of the header's path. This behavior helps finding the syntax-file if a package was renamed or is maintained with a different name locally by package authors than it will use later in Package Control.

Fixes #150.

### DIFF
--- a/plugins_/syntaxtest_dev.py
+++ b/plugins_/syntaxtest_dev.py
@@ -375,19 +375,22 @@ class PackagedevSuggestSyntaxTestCommand(sublime_plugin.TextCommand):
 
 
 class AssignSyntaxTestSyntaxListener(sublime_plugin.EventListener):
-
     """Assign target syntax highlighting to a syntax test file."""
 
     def on_load_async(self, view):
+        plain_text = "Packages/Text/Plain text.tmLanguage"
         test_header = get_syntax_test_tokens(view)
         if test_header and test_header.syntax_file:
-            if view.settings().get('syntax', None) != test_header.syntax_file:
-                syntax = test_header.syntax_file
-                *_, file_name = syntax.rpartition('/')
-                if syntax in sublime.find_resources(file_name):
-                    view.assign_syntax(syntax)
-                else:  # file doesn't exist
-                    view.assign_syntax("Packages/Text/Plain text.tmLanguage")
+            view_syntax = view.settings().get('syntax', plain_text)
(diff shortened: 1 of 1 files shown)


plugins_/syntaxtest_dev.py
```diff
@@ -375,19 +375,22 @@ def run(self, edit, character='^'):
...
         if test_header and test_header.syntax_file:
-            if view.settings().get('syntax', None) != test_header.syntax_file:
-                syntax = test_header.syntax_file
-                *_, file_name = syntax.rpartition('/')
-                if syntax in sublime.find_resources(file_name):
-                    view.assign_syntax(syntax)
-                else:  # file doesn't exist
-                    view.assign_syntax("Packages/Text/Plain text.tmLanguage")
+            view_syntax = view.settings().get('syntax', plain_text)
+            if view_syntax != test_header.syntax_file:
```
FichteFoll: The entire code block is fundamentally flawed, starting with this line. I'll just rewrite it.

FichteFoll:
Fixed in 46184970381a3f0a4866d4e4d276db8bb7d5c5dc.

PR #166, opened 2017-10-16
Initial syntax highlighting for the new sublime-color-scheme format

keith-hall:
...introduced in build 3149


FichteFoll:
Maybe you'd like to add a few completions as well?

keith-hall:
Added some static completions. Probably we will want to parse the `variables` section using a plugin to offer suggestions when people write `var(` in the relevant places, just like we do for `sublime-syntax` files.

keith-hall:
Although, thinking about it, maybe the buffer completions will be good enough?

FichteFoll:
We'll probably want some CSS4 color highlighting first, that we could then also use for a theme syntax that doesn't exit yet.

Package/Sublime Text Color Scheme/Completions/Globals Keys.sublime-completions
```diff
@@ -0,0 +1,31 @@
...
+        { "trigger": "shadow\tglobals key", "contents": "shadow" },
+        { "trigger": "invisibles\tglobals key", "contents": "invisibles" },
+        { "trigger": "caret\tglobals key", "contents": "caret" },
+        { "trigger": "line_highlight\tglobals key", "contents": "line_highlight" },
+        { "trigger": "foreground\tglobals key", "contents": "foreground" },
+        { "trigger": "background\tglobals key", "contents": "background" },
+        { "trigger": "selection\tglobals key", "contents": "selection" },
+        { "trigger": "selection_border\tglobals key", "contents": "selection_border" },
+        { "trigger": "inactive_selection\tglobals key", "contents": "inactive_selection" },
+        { "trigger": "highlight\tglobals key", "contents": "highlight" },
```
jrappen: there's also a `highlight_foreground`

Package/Sublime Text Color Scheme/Completions/Globals Keys.sublime-completions
```diff
@@ -0,0 +1,31 @@
+{
+    "scope": "meta.globals-key.sublime-color-scheme",
+    "completions": [
+        { "trigger": "accent\tglobals key", "contents": "accent" },
+        { "trigger": "shadow\tglobals key", "contents": "shadow" },
+        { "trigger": "invisibles\tglobals key", "contents": "invisibles" },
```
jrappen: there's also `rulers`

Package/Sublime Text Color Scheme/Completions/Globals Keys.sublime-completions
```diff
@@ -0,0 +1,31 @@
...
+        { "trigger": "misspelling\tglobals key", "contents": "misspelling" },
+        { "trigger": "brackets_foreground\tglobals key", "contents": "brackets_foreground" },
+        { "trigger": "tags_foreground\tglobals key", "contents": "tags_foreground" },
+        { "trigger": "shadow_width\tglobals key", "contents": "shadow_width" },
+        { "trigger": "bracket_options\tglobals key", "contents": "bracket_options" },
+        { "trigger": "bracket_contents_options\tglobals key", "contents": "bracket_contents_options" },
+        { "trigger": "tag_options\tglobals key", "contents": "tag_options" },
+        { "trigger": "popup_css\tglobals key", "contents": "popup_css" },
+        { "trigger": "phantom_css\tglobals key", "contents": "phantom_css" },
+        { "trigger": "implicit_selection_foreground\tglobals key", "contents": "implicit_selection_foreground" },
```
jrappen: ..., `"minimap_border": "..."`, `"gutter": "..."`, `"gutter_foreground": "..."`, ...

Package/Sublime Text Color Scheme/Completions/Globals Keys.sublime-completions
keith-hall: I can't see `highlightForeground` mentioned at https://www.sublimetext.com/docs/3/color_schemes.html, am I missing something?

PR #171, opened 2017-11-07
All leaf values in color scheme files must be strings

### LINKED ISSUE #170: Quoted bool value in .sublime-color-scheme is marked as invalid.
I have a quoted bool value in .sublime-color-scheme and PackageDev marks it as invalid.

image

However, if I unquote it, ST complains that it's invalid.
image

keith-hall:
Fixes #170


FichteFoll:
Yay for proper type usage.

PR #179, opened 2018-01-08
Add support for "keyfiles" to sublime-build syntax highlighting

keith-hall:
while looking at https://forum.sublimetext.com/t/trying-to-create-a-debian-package-sublime-build/34430, I noticed that `keyfiles` wasn't highlighted correctly. This PR corrects that and adds the relevant completions. More info at:
- http://www.sublimetext.com/docs/3/build_systems.html#options
- https://forum.sublimetext.com/t/custom-build-system-for-rails/24609/2
- https://forum.sublimetext.com/t/build-systems/14435
- https://forum.sublimetext.com/t/dev-build-3072/14605


keith-hall:
This now also fixes a bug whereby the `cancel` key in build systems was updated in recent dev builds, but the highlighting was scoping valid constructs as `invalid` and didn't recover gracefully either.

PR #188, opened 2018-05-12
Snippets: Highlight CDATA as self closing tag

deathaxe:
Only the encapsulated content of a <![CDATA[ content ]]> tag should be scoped as string. By scoping the beginning and the end as ordinary xml tags the readability can be improved. The content can better be distinguished from the enclosing tag.

The <![CDATA[ is parsed by the xml-parser and marks the start of text to exclude from parsing up to the ]]> tag.

This commit implements the required changes to do so and fixes a bug which causes the </content> tag after an illegal second CDATA tag to be highlighted as illegal as well.


FichteFoll:
Although I liked my `punctuation.definition.string` highlighting, this is obviously the more general way. I can always add `meta.tag.cdata` to my color scheme.

PR #190, opened 2018-06-14
add the resize_window command to the list of known, built-in commands

keith-hall:
as mentioned at https://github.com/SublimeTextIssues/Core/issues/242#issuecomment-397289860


keith-hall:
I've also updated the PR to include some missing commands, and unify the `command_type`s (`view` -> `text`) to match the [expected functionality](https://github.com/SublimeText/PackageDev/blob/dad145a495b5228bcad01eaa6159e9ceccc0d8f7/plugins_/command_completions/commandinfo.py#L79).

PR #192, opened 2018-06-29
validate built-in command  metadata

keith-hall:
This PR ensures that when the PackageDev `command_completions` plugin is loaded, it will check that the metadata could be loaded successfully, and warn if it couldn't. It will also log a warning if the metadata file, which is supposed to contain built-in (core) commands only, contains a command defined in the `Default` package. This will allow PackageDev developers to make changes to the metadata file, restart ST and confirm (from looking at the ST console) that the file is still parse-able and doesn't contain any mistakes. It also highlights that I shouldn't have added the indentation related commands in https://github.com/SublimeText/PackageDev/pull/190, as I forgot they exist in the `Default` package. (For some reason they weren't showing up as completions when I was trying something out the other day, which is what inspired me to add them.)

@FichteFoll, if we don't want it to parse the metadata on plugin load, and instead defer it to later - when the file is actually read through normal use of the plugin, that's fine by me - my main goal here is to have a way to confirm that the metadata file is correct, and if it doesn't validate automatically but I can still execute the verification routine manually on demand, then that is fine by me.

Unfortunately I don't think we'll be able to set up the CI to spot check this for PRs, but if you can think of a way that'd be great! ;)

### DIFF
--- a/plugins_/command_completions/__init__.py
+++ b/plugins_/command_completions/__init__.py
@@ -13,7 +13,8 @@ from .commandinfo import (
     get_builtin_command_meta_data,
     get_builtin_commands,
     iter_python_command_classes,
-    get_args_from_command_name
+    get_args_from_command_name,
+    validate_commands
 )
 
 __all__ = (
@@ -247,3 +248,7 @@ class SublimeTextCommandArgsCompletionPythonListener(sublime_plugin.EventListene
 
         completions = [("args\tauto-detected arguments", args)]
         return completions
+
+
+def plugin_loaded():
+    validate_commands()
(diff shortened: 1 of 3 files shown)


FichteFoll:
I think I'd rather validate within `get_builtin_commands`, which is guarded by a `lru_cache` so lazily executed, with the option to run it on-demand (by importing the function from the console, for example - like I do with [`_check_missing`](https://github.com/SublimeText/PackageDev/blob/d47794a658cbeb5a541ce47d02a32578dbd66cd8/plugins_/__init__.py#L30)).

FichteFoll:
Also, I didn't check, but the reason for adding the commands to the build-ins manually might also be to provide a docstring. Though, I don't remember whether we even use that anywhere.

keith-hall:
That's a good point - this seems to be the first instance of us including a `Default` package's command in this meta data file, which is why I figured it was not something we normally do. I also don't remember if we use the docstring anywhere, but I haven't seen it used. Maybe in the future we could add a tooltip when hovering over a `.run_command(` method and over the `command` in the keymap etc. to show the docstring.

PR #194, opened 2018-09-18
Fix: sublime-color-scheme main key completions

deathaxe:
The main-keys of a sublime-color-scheme are currently being suggested within themselves.

This commit changes the selector to suggest main keys only within the top-level object where such keys are actually valid only.


FichteFoll:
Uh, the `- meta.mapping meta.mapping` should already prevent this from happeneing. I'll have to take a closer look why it's not working as intended as it is now.

deathaxe:
Hmm. Maybe `- (meta.mapping meta.mapping)` ?

deathaxe:
@FichteFoll: How about main keys being suggested within themselves? I recently simplified my initial approach to fix that issue.

FichteFoll:
Ah yeah, I forgot about this. I'll just include it in the next hotfix release.

FichteFoll:
Made a small modification and went with the same approach as with themes.

PR #195, opened 2018-10-08
Initial sublime-theme syntax definition added

keith-hall:
This is just the basics at the moment, copied from the color scheme. We will also want proper scoping of all known theme rule keys (and values) and autocompletion eventually, but this at least makes working with variables easier for now (in new the theme format).

### DIFF
--- a/Package/Sublime Text Theme/Sublime Text Theme.sublime-syntax
+++ b/Package/Sublime Text Theme/Sublime Text Theme.sublime-syntax
@@ -1,0 +1,216 @@
+%YAML 1.2
+---
+# http://www.sublimetext.com/docs/3/syntax.html
+name: Sublime Text Theme (JSON)
+file_extensions:
+  - sublime-theme
+scope: source.json.sublime.theme
+
+contexts:
+  main:
+    - match: (?=\{)
+      push: main-mapping
+    - include: rules-sequence-pop
+    - include: expect-mapping-rest
+
+  main-mapping:
+    - match: \{
+      scope: punctuation.section.mapping.begin.json
+      push:
+        - meta_scope: meta.theme.collection.sublime-theme meta.mapping.json
+        - match: \}
+          scope: punctuation.section.mapping.end.json
+          set: only-comments
+        - match: (?=")
+          push: [in-mapping-expect-comma, main-key]
+        - include: expect-key-rest
+
+  main-key:
+    - clear_scopes: 1
+    - match: (")(rules)(")
+      scope: meta.mapping.key.json meta.main-key.sublime-theme string.quoted.double.json
+      captures:
+        1: punctuation.definition.string.begin.json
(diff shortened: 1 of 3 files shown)


Package/Sublime Text Theme/Sublime Text Theme.sublime-syntax
```diff
@@ -0,0 +1,216 @@
...
+        3: punctuation.definition.string.end.json
+      set: [expect-rules-sequence-value, expect-colon]
+    - match: (")(variables)(")
+      scope: meta.mapping.key.json meta.main-key.sublime-theme string.quoted.double.json
+      captures:
+        1: punctuation.definition.string.begin.json
+        2: keyword.other.main.sublime-theme
+        3: punctuation.definition.string.end.json
+      set: [expect-variables-mapping-value, expect-colon]
+    - match: (")(name|author)(")
```
FichteFoll: Do themes really have an author field?
I mean, what does it do? If it has no special purpose, I'd rather have generic "doesn't do anything" highlighting than a `keyword` scope.

Same for the name key, actually.

Package/Sublime Text Theme/Sublime Text Theme.sublime-syntax
```diff
@@ -0,0 +1,216 @@
...
+      scope: meta.mapping.key.json meta.rule-key.class.sublime-theme string.quoted.double.json
+      captures:
+        1: punctuation.definition.string.begin.json
+        2: keyword.other.rule.sublime-theme
+        3: punctuation.definition.string.end.json
+      set: [expect-string-value, expect-colon] # TODO: mark the string value as something to be indexed?
+    # TODO: layerx.etc
+    # TODO: other rule keys and values
+    - include: in-dictionary-rule-key
+    - include: expect-mapping-rest
```
FichteFoll: `expect-key-rest`

Package/Sublime Text Theme/Sublime Text Theme.sublime-syntax
```diff
@@ -0,0 +1,216 @@
...
+        - match: (?=")
+          set:
+            - match: '"'
+              scope: string.quoted.double.json meta.variable-name.sublime-theme punctuation.definition.string.end.json
+              set: [expect-color-string-value-or-json-value, expect-colon]
+        - match: $\n?
+          scope: invalid.illegal.unclosed-string.json
+          set: [expect-color-string-value, expect-colon]
+        - include: string-escape
+    - include: in-dictionary-main-key
```
FichteFoll: This will set `meta.main-key.sublime-theme`, which isn't what we want.

FichteFoll:
I'll merge for now, but this will need some more attention before the next release.

FichteFoll:
FYI, I added highlighting for a couple more keys and made the TODO comments more specific (after going through the entire doc page and some default themes).

PR #202, opened 2018-11-20
Extend carets further in syntax test completions.

Thom1729:
When extending assertions in syntax tests, keep extending the assertion until the asserted scope would change.

Given (curly brackets denote selection):

```js

  [a, b, c, d]
//^{ meta.sequence punctuation.section.brackets.begin}
```

When you type a caret, the old result was:

```js

  [a, b, c, d]
//^^{ meta.sequence}
```

And the new result is:

```js

  [a, b, c, d]
//^^^^^^^^^^^^{ meta.sequence}
```


FichteFoll:
I've always wanted this, but was too lazy to implement it whenever it bothered me until I eventually forgot about it again. Thanks.

PR #204, opened 2018-11-20
Remove old code

Thom1729:
Add a dependency on `sublime_lib` and remove a bunch of code that's no longer necessary:

- OutputPanel.
- Edit.
- The various Settings classes.
- Some unused utility functions.


plugins_/lib/fileconv/loaders.py
```diff
@@ -172,11 +174,17 @@ def __init__(self, window, view, file_path=None, output=None, *args, **kwargs):
...
             self.output = output
         else:
-            self.output = OutputPanel(self.window, self.output_panel_name,
-                                      file_regex=self.file_regex, path=path)
+            self.output = OutputPanel.create(
+                self.window,
+                self.output_panel_name
+            )
+
+        SettingsDict(self.output.view.settings()).update(
```
FichteFoll: This seems like overkill. Not really necessary to add a dependency compared to fetching the settings object and assigning the settings normally. It would even be less lines.

Thom1729: Done.

Thom1729:
I agree that we should rename the stuff that remains, but I don't think we need to do that in this PR. There is definitely more that can be removed, and I wouldn't be surprised if some functionality were added to `sublime_lib` as a result of this refactoring, so it's hard to say what the remnants will look like just yet.

Plus, this is already touching a lot of lines in a lot of files.

FichteFoll:
I just believe that a state in which there are two `sublime_lib`s is undesireable.

But don't worry, I'll go ahead and do it myself.

PR #209, opened 2019-01-04
Fix Context completions in multi-push statements not displayed

### LINKED ISSUE #201: Context completions in multi-push statements
Currently context auto-completion works only, if the list of contexts is surrounded by brackets.

```
  cmd-generic:
    - match: (?=\S)
      push: [cmd-generic-end, cmd-generic-name]
```

It would be useful to have auto-competions for contexts like `cmd-generic-end` in following list styles as well.

```
  cmd-generic:
    - match: (?=\S)
      push: 
        - cmd-generic-end
        - cmd-generic-name
```

deathaxe:
Fixes #201

This commit fixes an issue with multi-push/multi-set contexts not being scoped as include-list, if the sequence spans multiple lines.

### DIFF
--- a/Package/Sublime Text Syntax Definition/Sublime Text Syntax Definition.sublime-syntax
+++ b/Package/Sublime Text Syntax Definition/Sublime Text Syntax Definition.sublime-syntax
@@ -378,7 +378,7 @@ contexts:
               scope: punctuation.separator.context-name.sublime-syntax
               set: expect_include
         # absolute includes
-        - match: \w((?!\s#|:\s).)+\.sublime-syntax(?=\s*(#|$))
+        - match: \w((?!\s#|:\s)[^{{c_flow_indicator}}])+\.sublime-syntax(?=\s*(#|$))
           scope: support.module.file-path.sublime-syntax
           set:
             - match: '{{_flow_scalar_end_plain_in}}'
--- a/Package/Sublime Text Syntax Definition/Sublime Text Syntax Definition.sublime-syntax
+++ b/Package/Sublime Text Syntax Definition/Sublime Text Syntax Definition.sublime-syntax
@@ -316,7 +316,6 @@ contexts:
       pop: true
 
   expect_include_list:
-    # only flow list for now, or plain string
     - meta_content_scope: meta.expect-include-list
     - match: \[
       scope: punctuation.definition.array.begin.sublime-syntax
@@ -330,16 +329,34 @@ contexts:
         - include: comment
         - include: yaml-tags-anchors
         - include: include
-    - match: \-(?=\s*{{plain_scalar_but_not_block_key}})
-      scope: punctuation.definition.block.sequence.yaml
-      push:
-        - meta_scope: meta.include-list.sublime-syntax meta.flow-sequence.yaml
-        - match: $\n?
-          pop: true
+    - match: ^
+      set:
+        - meta_scope: meta.expect-include-list-or-content
         - include: comment
-        - include: yaml-tags-anchors
-        - include: include
+        - include: yaml-block-sequence
+        # first item after push/set looks like an included context
+        - match: ^([ ]+)(?=-\s*{{plain_scalar_but_not_block_key}})
+          set:
+            - meta_content_scope: meta.expect-include-list-or-content meta.include-list.sublime-syntax
+            - include: include_list_content
+            - match: ^(?=([ ]+)-)
+              set:
+                - meta_scope: meta.include-list.sublime-syntax
+                - include: include_list_content
+        # limit context to the current line
+        - match: $|(?=\S)
+          pop: true
     - include: expect_include
+
+  include_list_content:
+    - include: comment
+    - include: include
+    - include: yaml-block-sequence
+    # pop off at none-empty line with different indention than first include item
+    - match: ^(?!(\s*$|\1-))
+      pop: true
+    - match: \S.+$
+      scope: invalid.illegal.include.sublime-syntax
 
   include:
     - match: '{{plain_scalar_but_not_block_key}}'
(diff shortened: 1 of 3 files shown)


Package/Sublime Text Syntax Definition/Sublime Text Syntax Definition.sublime-syntax
```diff
@@ -330,6 +330,15 @@ contexts:
         - include: comment
         - include: yaml-tags-anchors
         - include: include
+    - match: \-(?=\s*{{plain_scalar_but_not_block_key}})
+      scope: punctuation.definition.block.sequence.yaml
+      push:
+        - meta_scope: meta.include-list.sublime-syntax meta.flow-sequence.yaml
+        - match: $\n?
```
FichteFoll: Capturing the newline will cause it to also have the meta scope applied, which seems weird to me considering the context won't span the next line.
Instead, only match on the anchor.

```suggestion
        - match: $
```

Alternatively, you could capture the indentation of the first item and only pop the context on a line with less indentation (whereas the `-` counts towards the indentation), but I don't think that's necessary.

Package/Sublime Text Syntax Definition/syntax_test_sublime-syntax.yaml
```diff
@@ -264,6 +264,23 @@ contexts: !mytag
 #          ^^^^^^ storage.type.tag-handle.yaml
 #                 ^^^^ constant.language.boolean.yaml
 
+    - match: foo
+#   ^ meta.block.contexts.sublime-syntax punctuation.definition.block.sequence.item.yaml
+#     ^^^^^ meta.block.contexts.sublime-syntax string.unquoted.plain.out.yaml keyword.other.match.sublime-syntax
+      push:
+#     ^^^^ meta.block.contexts.sublime-syntax string.unquoted.plain.out.yaml keyword.control.flow.push.sublime-syntax
+        - context-name-3
+#       ^^^^^^^^^^^^^^^^^ meta.block.contexts.sublime-syntax meta.expect-include-list meta.include-list.sublime-syntax meta.flow-sequence.yaml
```
FichteFoll: Which will then require the tests to be updated as follows:

```suggestion
#       ^^^^^^^^^^^^^^^^ meta.block.contexts.sublime-syntax meta.expect-include-list meta.include-list.sublime-syntax meta.flow-sequence.yaml
```

Package/Sublime Text Syntax Definition/syntax_test_sublime-syntax.yaml
```diff
@@ -264,6 +264,23 @@ contexts: !mytag
...
+#       ^ punctuation.definition.block.sequence.yaml
+#         ^^^^^^^^^^^^^^ meta.include.sublime-syntax string.unquoted.plain.out.yaml variable.other.sublime-syntax
+        - context-name-2
+#       ^^^^^^^^^^^^^^^^^ meta.block.contexts.sublime-syntax meta.expect-include-list meta.include-list.sublime-syntax meta.flow-sequence.yaml
```
FichteFoll:
```suggestion
#       ^^^^^^^^^^^^^^^^ meta.block.contexts.sublime-syntax meta.expect-include-list meta.include-list.sublime-syntax meta.flow-sequence.yaml
```

Package/Sublime Text Syntax Definition/Sublime Text Syntax Definition.sublime-syntax
```diff
@@ -330,6 +330,15 @@ contexts:
...
+          pop: true
+        - include: comment
+        - include: yaml-tags-anchors
+        - include: include
     - include: expect_include
```
FichteFoll: We don't seem to match tags here, do we? Mind verifying?

Package/Sublime Text Syntax Definition/Sublime Text Syntax Definition.sublime-syntax
deathaxe: The `\n`  is captured because I saw context not to be suggested, if the cursor was at the end of the line. After a double check, I seem to be wrong.

While working with the change for a while, I found the normal `keywords` not to be part of the completions anymore, if the caret is at the first line of the sequence.

```
- match: test
  push:
    - m|  <-- no match suggested
```

Need to look into it as well.

Package/Sublime Text Syntax Definition/Sublime Text Syntax Definition.sublime-syntax
deathaxe: Added it as it was part of the array above as well. Doesn't look like as we would need it.

deathaxe: Removed it

Package/Sublime Text Syntax Definition/Sublime Text Syntax Definition.sublime-syntax
deathaxe: Use the "Alternative solution" with a dedicated context.

Package/Sublime Text Syntax Definition/Sublime Text Syntax Definition.sublime-syntax
```diff
@@ -330,16 +329,35 @@ contexts:
...
+          pop: true
+    - match: (?=\S)
+      set: expect_include
+
+  include_list_content:
+    - include: comment
+    - include: include
+    - include: yaml-block-sequence
+    # pop off at none-empty line with different indention than first include item
+    - match: ^(?!(\s*$|\1-))
```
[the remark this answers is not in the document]
deathaxe: What exactly do you have in mind? Comment about the content of the context? Existing comment not sufficient enough?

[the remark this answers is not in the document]
FichteFoll: Ok, so this works because the YAML comment context starts matching at the beginning of the line, so it can actually match before this pop match.

PR #212, opened 2019-01-11
Add some input helpers for sublime-syntax files

deathaxe:
This commit adds some key bindings to:

1. auto-insert a newline, when typing a `:` after a context key. Context keys are recognized by being indented by at most two characters and not starting with `- `.
2. auto-insert a space after the first typed `:` in a line. This rule applies to situations when rules are inserted manually (not by auto-completion) or correcting them.
3. auto-insert space and trigger auto-completion, when typing the first `-` in a line, which indicates a new rule to be added. Up to this commit typing `-` directly triggers auto-completion as `-` was removed from word-separators. Committing the completion results in malformed yaml (e.g. -match: '').

Furthermore this commit adds a syntax-specific setting to ensure to trim trailing spaces, if the sublime-syntax file is saved.

### DIFF
--- a/Package/Default.sublime-keymap
+++ b/Package/Default.sublime-keymap
@@ -22,6 +22,28 @@
             { "key": "selector", "operator": "equal", "operand": "meta.scope.sublime-syntax" }
         ]
     },
+    {   // auto-insert space and trigger auto completion after hyphon (rules)
+        "keys": ["-"], "command": "run_macro_file", "args": {"file": "res://Packages/PackageDev/Package/Sublime Text Syntax Definition/Add Syntax Rule.sublime-macro"}, "context": [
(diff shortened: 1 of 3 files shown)


FichteFoll:
I like 3., but I feel like the other two might be too intrusive and maybe unexpected. I'll have to play around with that first.

FichteFoll:
You get used to auto-insertion after `:` fast enough, so this isn't as intrusive as it seems. However, with the required fix for `- `, this means it's overall more consistent.

Fixed an issue with contexts not being completed for the first line with 832c2cfa16f042c8bc4fcaaa58ed6857bf3091d8.
Also made `captures:` insert a newline in 2e7f416731d0166bad41cc8dd2fe14d44985b16d.

PR #215, opened 2019-01-15
Revert "Fix context completions in multi-line push rules"

deathaxe:
This reverts commit 832c2cfa16f042c8bc4fcaaa58ed6857bf3091d8.

This commit probposes to revert 832c2c as it breaks command completions in the first line after a push/set statement.

The first line after a push/set can either be
- the start of a list of included contexts
- or it can contain normal commands like `- match`.

Hence auto-completion must suggest the merged list of available contexts and the normal command completions if the cursor is in the first line.

This does not work in v3.2.2 due to the mentioned commit.

We only see context completions now.

### DIFF
--- a/Package/Sublime Text Syntax Definition/Sublime Text Syntax Definition.sublime-syntax
+++ b/Package/Sublime Text Syntax Definition/Sublime Text Syntax Definition.sublime-syntax
@@ -335,8 +335,14 @@ contexts:
         - include: comment
         - include: yaml-block-sequence
         # first item after push/set looks like an included context
-        - match: ^([ ]+)(?=-\s*(?:{{plain_scalar_but_not_block_key}}|\s*$))
-          set: include_list_content
+        - match: ^([ ]+)(?=-\s*{{plain_scalar_but_not_block_key}})
+          set:
+            - meta_content_scope: meta.expect-include-list-or-content meta.include-list.sublime-syntax
+            - include: include_list_content
+            - match: ^(?=([ ]+)-)
+              set:
+                - meta_scope: meta.include-list.sublime-syntax
+                - include: include_list_content
         # limit context to the current line
         - match: $|(?=\S)
           pop: true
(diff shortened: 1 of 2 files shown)


FichteFoll:
While I get what you were going with there, I find it hardly useful to suggest key completions for the first list entry because it is *very* unlikely that you are just writing a `push` or `set` declaration and then leave it empty. Instead, it is much more likely that you want to insert a context and all the key completions get in your way. That was my impression when experimenting with it, at least.
This is the only case where this matters. Any other form of include list or a single context will terminate the include scope.

deathaxe:
You mean writing 

```YAML
- match foo
  push:
    - ma   # <- auto-completion suggesting "match:" or "machine-context"
```

is an unusual use-case?

In my world, the first line after `push:` can either contain a list of included contexts or normal commands like `- match`, `- meta_scope`, ... . Therefore I want to see both suggestions in the first line.

Just replacing command completions by context completions after push/set is no step forward, IMHO.

The content of this line then decides, whether we need to suggest contexts or keywords in the following lines then:

Context of keywords/commands

```YAML
- match foo
  push:
    - match: bar
      scope: keyword
``

versus

Context of included contexts

```YAML
- match foo
  push:
    - context-1
    - context-2
``

deathaxe:
1. This is not the question! It is a valid style of listing and hence should be well supported.
2. It comes in the way, if either names or number of contexts are large.

FichteFoll:
Ah no, you're right of course. I'll just blame it on being 3am at the time and not thinking of trying this with actual syntax definitions instead of testing in the test file. I'll revert it when I'm home again.

deathaxe:
No problem. I was in exactly the same situation not to see that issue. It was actually the reason for changing the `syntax_dev.py` to be able to return a merged list of context/keyword completions.

FichteFoll:
I'll make sure to add a comment somewhere so this doesn't happen again.

PR #222, opened 2019-02-01
Keymap: Add Goto Next/Prev Modification

deathaxe:
New builtin commands added with ST3189.


FichteFoll:
The command to trim trailing whitespace has also been converted to c++ code. Can you check whether the command still exists or whether the code now just operates on save?

I seem to remember that it was a command. Not at a computer.

FichteFoll:
>Did not find them in the yml file anyway.

Yes, since it was a Python command before, so it was getting detected automatically. I was asking in case we had to add it to the yaml file.

Thanks for clarifying the the commit message. :rocket:

Issue #219, opened 2019-01-19
New theme elements


FichteFoll:
I added the new attributes in #210 because they are needed for proper highlighting, but I haven't added the new elements because I'm waiting on an update of the official docs for those.

https://www.sublimetext.com/docs/3/dev/themes.html

I might also need to tweak the CSS `color` function since ST added a few custom modifier functions that the CSS4 draft does not have. Most likely need to implement my own version of that instead of relying on the shipped CSS.sublime-syntax. Update: fixed by #229

FichteFoll:
@deathaxe considering that you added many new theme properties, attributes, etc recently, do you happen to know where we stand with out coverage currently?

deathaxe:
Not for sure. Just added features/fixed stuff I realized in the Default.sublime-them while working on DA UI successor. Did not walk through the latest docs in detail.

FichteFoll:
Went through the changelog on the docs and added all things still missing in 8411ab6825f82e63ee8a5a337e75fa20f9d4d3e3 and 351a4c9b197e73126402d96ba767d6e9499a495f.

PR #223, opened 2019-02-02
Theme: Add status bar VCS indicator elements

deathaxe:
According to a recent forum post the following theme elements are now supported to style the new vcs status bar indicators.


FichteFoll:
Related: #219

PR #227, opened 2019-02-16
Theme/Scheme: Variables can contain hyphens

deathaxe:
The hyphen is removed from the list of word separators to fix goto definition/reference for variables, whose names contain them.

### DIFF
--- a/Package/Sublime Text Color Scheme/Sublime Text Color Scheme.sublime-settings
+++ b/Package/Sublime Text Color Scheme/Sublime Text Color Scheme.sublime-settings
@@ -4,4 +4,6 @@
     [
         {"characters": ".", "selector": "meta.scope-selector.sublime"}
     ],
+    "trim_trailing_white_space_on_save": true,
+    "word_separators": "./\\()\"':,.;<>~!@#$%^&*|+=[]{}`~?" // default without '-'
 }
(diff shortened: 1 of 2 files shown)


Package/Sublime Text Color Scheme/Sublime Text Color Scheme.sublime-settings
```diff
@@ -4,4 +4,6 @@
     [
         {"characters": ".", "selector": "meta.scope-selector.sublime"}
     ],
+    "trim_trailing_white_space_on_save": true,
```
FichteFoll: Why add this setting? It's not mandated by the syntax/format and doesn't have any benefit (other than generally being the correct default, but that's not for me to decide).

Same for the theme.

deathaxe: https://github.com/SublimeText/PackageDev/blob/7f470db736a40e4789f954dd7ee555b34d91d4cb/Package/Sublime%20Text%20Syntax%20Definition/Sublime%20Text%20Syntax%20Definition.sublime-settings#L14-L15

FichteFoll: Syntax definitions are written in YAML, not JSON, where trailing whitespace is ignored (except inside block scalars) and that's not immediately obvious to users. Trimming trailing whitespace *can* help prevent anti-patterns here, I guess.

Though you do have a point in that this *probably* doesn't really bring much to the table either. I'd rather remove whitespace trimming there as well than introducing it to the other syntaxes, however.

PR #228, opened 2019-02-16
Theme/Scheme: Add symbol/reference lists for variables

deathaxe:
This commit enables Goto Definition/Goto Reference for variables in

- sublime-color-schemes
- sublime-themes


FichteFoll:
Variables cannot be cross-referenced, which is why I chose not to add them to the indexed symbols. However, unfortunately Goto Definition depends on this, so I guess we do need it after all.

PR #235, opened 2019-03-03
Theme: Add more VCS status attributes

deathaxe:
This commit

1. adds `missing` and `deleted` vcs status attribute values
2. sorts all values according to the theming guideline

see: https://www.sublimetext.com/docs/3/dev/themes.html -> file_system_entry -> attributes


deathaxe:
@FichteFoll - what a timing :-)

FichteFoll:
I smell a release around the corner :wink:

PR #236, opened 2019-03-03
Theme: This commit adds the 'case' property

deathaxe:
The `case` property enables themes to modify the rendering of the side bar headings like `Folders` or `Open Files`.

It is implemented as string value by this commit.

As this property is basically an enumeration of "upper", "lower", "title", it should get its own meta scope and a value completion in the future.

The current implementation of the JSON/theme syntax mixture causes all additional meta scopes to be cleared from a string value. A new approach/implementation of string values is needed first to enable completions.


FichteFoll:
For context, implemented in https://github.com/SublimeText/PackageDev/commit/8411ab6825f82e63ee8a5a337e75fa20f9d4d3e3.

PR #250, opened 2019-07-04
Project: Fix Build Systems mapping

deathaxe:
The `build_systems` dictionary in a sublime-project file does not
contain a build system variant, but a list of full featured build
systems.

This means `selector` is a valid key. Additionally the `name` key is
used to identify the build system.

Note:

  The completions selectors might need to be tweaked a bit as some
  glitches were encountered. main keys are suggested within the `cmd`
  value part for instance. Looks like another issue.


FichteFoll:
Thanks for the PR and the report. Fixed the completions issues (minus the exception mentioned in the commit message) in 09b91fcfd62a1a96d5c214fc43c4640409ee305d.

PR #253, opened 2019-08-26
Update scope used in Keymap syntax to follow conventions

seanCodes:
Per the [scope naming conventions for `string.`](https://www.sublimetext.com/docs/3/scope_naming.html#string), open quotes should use the scope `punctuation.definition.string.begin` instead of `punctuation.definition.string.start`.


FichteFoll:
Ah, yes, good catch. `begin` is also used everywhere else, so this is obviously an oversight.

PR #258, opened 2019-09-19
Settings: Improve completions of known values

deathaxe:
Had a look at  8b904c6 and found some possible edge cases to fix.

While looking into it, I found the `(default)` marker not to be added to `color_scheme` and `theme` completions. Fixed it by extending the `format_completion_item()`. 

Finally found the `color_scheme` completions missing the `sublime-color-scheme` files and fixed it.

_Note:_

The extended `format_completion_item()` could be used in `_completions_from_comment()` and `_completions_from_default()` to add the `(default)` mark, which would result in `_marked_default_completions()` not being needed anymore. Left that for now as I am not sure whether it was the better alternative.

The main reason for extending `format_completion_item()` was to avoid breaking the description part of the completion label by calling `_marked_default_completions()`  on color_scheme, encoding or theme completions.

### DIFF
--- a/plugins_/settings/known_settings.py
+++ b/plugins_/settings/known_settings.py
@@ -578,11 +578,11 @@ class KnownSettings(object):
         l.debug("default value: %r", default)
 
         if key == 'color_scheme':
-            completions = self._color_scheme_completions()
+            completions = self._color_scheme_completions(default)
         elif key in ('default_encoding', 'fallback_encoding'):
             completions = self._encoding_completions(default)
         elif key == 'theme':
-            completions = self._theme_completions()
+            completions = self._theme_completions(default)
         else:
             completions = self._completions_from_comment(key)
             completions |= self._completions_from_default(key, default)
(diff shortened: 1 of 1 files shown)


plugins_/settings/known_settings.py
```diff
@@ -708,12 +708,18 @@ def _color_scheme_completions(default):
...
+                    value=name, is_default=name == default, description=package))
+
         for scheme_path in sublime.find_resources("*.tmTheme"):
             if not any(hide in scheme_path for hide in hidden):
-                _, package, *_, file_name = scheme_path.split("/")
+                _, package, *_, name = scheme_path.split("/")
                 completions.add(format_completion_item(
                     value=scheme_path, is_default=scheme_path == default,
-                    label=file_name, description=package))
+                    label=name, description=package))
```
FichteFoll: Now that I see this, tmTheme files need to be specified by path, not by name. I'll fix this.

PR #282, opened 2020-02-24
add scope completion for punctuation.definition.numeric.base

keith-hall:
This adds `punctuation.definition.numeric.base` to the scope data, so that it will be suggested in the auto completions when writing syntax definitions. (This was recently standardized, see i.e. https://github.com/sublimehq/Packages/pull/2175)


FichteFoll:
Note to self: Was pulled from the release via #284.

PR #283, opened 2020-02-25
fix bug preventing whitespace in the syntax test token

keith-hall:
ST supports whitespace in the middle of the syntax test token, but PackageDev wasn't offering any syntax test completions. This fixes that oversight to bring PackageDev back in line with ST behavior.

### DIFF
--- a/plugins_/syntaxtest_dev.py
+++ b/plugins_/syntaxtest_dev.py
@@ -42,7 +42,7 @@ def get_syntax_test_tokens(view):
     match = None
     if line.size() < 1000:  # no point checking longer lines as they are unlikely to match
         first_line = view.substr(line)
-        match = re.match(r'^(?P<comment_start>\s*\S+)'
+        match = re.match(r'^(?P<comment_start>\s*.+)'
                          r'\s+SYNTAX TEST\s+'
                          r'"(?P<syntax_file>[^"]+)"'
                          r'\s*(?P<comment_end>\S+)?$', first_line)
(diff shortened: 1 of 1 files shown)


plugins_/syntaxtest_dev.py
```diff
@@ -42,7 +42,7 @@ def get_syntax_test_tokens(view):
     match = None
     if line.size() < 1000:  # no point checking longer lines as they are unlikely to match
         first_line = view.substr(line)
-        match = re.match(r'^(?P<comment_start>\s*\S+)'
+        match = re.match(r'^(?P<comment_start>\s*.+)'
                          r'\s+SYNTAX TEST\s+'
```
FichteFoll: This regex will match extraneous whitespace (any more than one) between the syntax test marker and the `SYNTAX TEST` token as part of the marker. Does this represent ST's internal behavior or should the any match be made non-greedy? (i.e. `\s*.+?`)

FichteFoll:
Just out of curiousity, what did you need this for?

plugins_/syntaxtest_dev.py
keith-hall: Good question, I will investigate that a bit more and get back to you, thanks!

keith-hall:
I am working on a syntax definition for LDraw files, for which the comment marker is `0 //` - the zero followed by whitespace represents a "meta command" and the double slashes ensure that the comment is disambiguated from a custom/unofficial command

plugins_/syntaxtest_dev.py
keith-hall: you were correct - it should be non-greedy. I have pushed a new commit to fix this.

PR #288, opened 2020-04-07
Update XML hidden_extensions hotfix for latest ST3

frou:
For those of us who started using PackageDev from [this time](https://github.com/sublimehq/Packages/commit/4a3712b7e236f8c4b443282d97bad17f68df318c) onwards, the magic in this plugin will never have been run.

### DIFF
--- a/plugins_/xml_hidden_extensions_hotfix.py
+++ b/plugins_/xml_hidden_extensions_hotfix.py
@@ -15,12 +15,12 @@ import sublime
 
 
 DEFAULT_VALUE = ["rss", "sublime-snippet", "vcproj", "tmLanguage", "tmTheme", "tmSnippet",
-                 "tmPreferences", "dae"]
-MODIFIED_VALUE = ["rss", "vcproj", "tmLanguage", "tmTheme", "tmSnippet", "dae"]
+                 "tmPreferences", "dae", "csproj"]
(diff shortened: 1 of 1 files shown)


FichteFoll:
Thanks, I totally forgot that I hardcoded the expected default value in that function to not unexpectedly break things. While this allows for safety in the manner that we only operate on known values, it also means it's not futureproof at all. I was hoping this gimmicky setting would never be modified again, considering my reports about it. However, I don't want to check whether the underlying value has changed on every plugin load either.

While your change makes it work for the latest version, I'd be more interesting in rewriting this in a way that allows for the underlying setting to change and our override to adapt to that. I have an idea for that going forward and will work on it on my next "release sprint".

Probably something to the likes of erasing the setting if the bootstrapped value exists and has the value I used previously and then adjust the setting by removing the extensions we don't want to be in there instead of hardcoding the before and after state. That way I only have to update the bootstrap value in the event the setting changes again.

frou:
In my personal `User/XML.sublime-settings` I have just put the following. Because the miscellaneous stuff that the stock `hidden_extensions` is trying to help with (RSS? Visual Studio metadata?), I don't care about.

```
"hidden_extensions": [],

"extensions": [
	"tmLanguage",
	"tmTheme",
	"tmSnippet",
]
```

Should I close this PR?

FichteFoll:
I would only want to close it if it's either fixed or a replacement issue was created. Unless you care about the number of open PRs, let's use this for tracking the issue.

PR #294, opened 2020-05-10
Add settings to disable tooltips and auto complete in settings

### LINKED ISSUE #286: Allow disabling completions/tooltips in settings
The new version of https://packagecontrol.io/packages/LSP-json (not released at the time of writing) will provide schema validation and autocomplete for settings through json schemas. With that enabled, autocomplete values for settings end up being duplicated. It would be nice to have an option to disable PackageDev ability to provide completions.

Screenshot 2020-03-20 at 09 08 07

rchl:
Resolves #286

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *.pyc
 *.cache
+*.mypy_cache
 *.sublime-project
 *.sublime-workspace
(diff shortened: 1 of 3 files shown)


rwols:
@FichteFoll any change you can release 3.2.17 for ST4 as well?

FichteFoll:
Yes and no. Yes, it would be possible, but I would really just get 3.3 out instead. I only need to find some time to finalize it.

PR #296, opened 2020-05-26
Syntax: Add highlighting for section comments

deathaxe:
It's not an official feature of a sublime-syntax file, but as I started to split syntaxes into sections for better readability I'd find it a good idea to also use them for navigation.

Syntaxes which make use of sections are:

- HTML (ASP) (master)
- Bash (master)
- CSS (master)
- Erlang (master)
- Haskell (see: https://github.com/sublimehq/Packages/pull/2679)
- HTML (master)
- Java (see: https://github.com/sublimehq/Packages/pull/2654)
- Markdown (some in master)
- Matlab (see: https://github.com/sublimehq/Packages/pull/2650)
- Perl (master)
- Ruby (some in master)
- XML (master)

Inspired by the contexts recently added to R.
see: https://github.com/sublimehq/Packages/commit/d43b3b7f92d306cfad6cac8e0c0574ef5f4afff5

### DIFF
--- a/Package/Sublime Text Syntax Definition/Sublime Text Syntax Definition.sublime-syntax
+++ b/Package/Sublime Text Syntax Definition/Sublime Text Syntax Definition.sublime-syntax
@@ -87,6 +87,13 @@ variables:
 
 contexts:
   comment:
+    ###[ SECTION COMMENT ]######
+    - match: ^#+(\[)\s*([^#\]]+?)\s*(\])#+$
+      scope: comment.line.double-dash.syntax
+      captures:
+        1: punctuation.section.brackets.begin.syntax
+        2: entity.name.section.syntax
+        3: punctuation.section.brackets.end.syntax
     - include: Packages/YAML/YAML.sublime-syntax#comment
 
   main:
--- a/Package/Sublime Text Syntax Definition/Symbols - Section.tmPreferences
+++ b/Package/Sublime Text Syntax Definition/Symbols - Section.tmPreferences
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <key>scope</key>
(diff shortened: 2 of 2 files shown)


FichteFoll:
While I do see value in having some sort of section markers in a long file, this is very much non-standard. Many people use different syntaxes for denoting sections and I don't see a compelling reason for why we should support this syntax in particular.

deathaxe:
I decided to use this style of grouping related contexts in all my recent syntax updates, many of them merged and shipped with ST. 

I like the one-liner style and the ability to break long files into logical pieces. They allow folding related stuff which makes editing easier. 

Such a kind of delimiters is especially useful as inheritance requires as many contexts as possible to be turned into named contexts because an inherited syntax would otherwise not be able to provide overrides.

We could say: A syntax desined to be a base syntax for inheritance must not use anonymous contexts, which causes many named contexts to be added to a syntax file, which need to be grouped logically somehow.

FichteFoll:
> I decided to use this style of grouping related contexts in all my recent syntax updates, many of them merged and shipped with ST.

Could you point me to some examples? I'm not finding anything other than for R. https://github.com/sublimehq/Packages/search?q=entity.name.section

I'm not opposed to adding some sort of section marker to syntax highlighting and I'm okay with more or less dictating the style to be used, since I believe we have this freedom as the only package providing syntax definitions for ST resource files. 
But there are way too many different section marking comment styles out in the wild and I'd like the marker to be consistent with other (popular) syntaxes and even better if the default packages used the same schema. Has Will ever commented on this?

deathaxe:
Is the list in the initial post not enough?

FichteFoll:
Ah, I think I completely missed what meant with your examples. I was looking for matches in the syntax definitions that would apply `entity.name.section` to a specifically formatted comment instead of instances of what you mean to highlight with this PR.

I doubt we'll get a "wider adoption" of this standard other than the default packages, probably, and at this point I'm fine with dictating the format. I believe we'd want to add a snippet that inserts a comment formatted like this and document that in the wiki, but I can take care of that when I prepare another release.

deathaxe:
Even though not enforced by the syntax and some rules exceeding it anyway, I always tried to adjust those delimiters to 80 chars, just to make it look well, if everything is collapsed. Not sure if a snippet could also help with that as the captions are typed after committing it.

Issue #292, opened 2020-04-11
Use the new completion features


FichteFoll:
In order to support the new completion features, I'll likely make the next or the following build ST4-only. Very useful new features like #289 might get backported to an st3 branch, but I don't think there would need to be much maintenance for that branch otherwise. PD overall seems very stable with few bugs and most of the issues or PRs I get in are feature requests. I would then decide on a case-by-case basis if it's worth backporting.

For clarification, this includes `sublime-completions` files and plugin-provided. Especially because of the latter, I should probably do #289 first to have it backwards compatible.

FichteFoll:
As an update, this is *almost* done. What's missing are theme and command completions, which I hope to finish tomorrow and then roll out the new release.

FichteFoll:
Settings & command completions are done as well. Just theme (and scheme) left.

FichteFoll:
All completions should use this the new format now, both plugins and completions files. 

Unfortunatly, I wasn't able to inhibit buffer completions. You can see my attempt at 83de08a13c311fdb9631e971a16d2f4c0269af8b, but it doesn't work as expected. Moved that to https://github.com/sublimehq/sublime_text/issues/3643.

PR #301, opened 2020-07-25
Syntax: Add new keywords of ST 4077+

deathaxe:
This commit adds the new keywords of ST 4077

- extends
- version
- hidden_extensions
- include_prototype

The match rules in the main context are sorted logically as they are expected to appear in a syntax file. Missing keywords from older builds such as `name` are added as well.


FichteFoll:
Thanks, that means less work for me. Could you address the flake8 notice?

deathaxe:
Looks like your flake setup complains about more than my change.

FichteFoll:
Yes, but that's okay. I'll fix that myself.

deathaxe:
According to #292, this PR would go into a new ST4 only branch? Would propably not make sense to display them in ST3.

FichteFoll:
I haven't fully decided what I'm going to do with the next release yet, i.e. whether I will do an ST3 one or only focus on ST4. I intend to figure that out once I'm done with the issues I still want and need to tackle for the next ST4 release and then decide if I'm going to make an ST3 branch with cherry-picked changes. It's mostly going to be a decision based on how much work I expect it to be.

I wanted to have this done weeks ago, but for some reason I can't seem to get a couple hours free.

PR #303, opened 2020-07-26
Syntax: Update input helper

deathaxe:
This PR proposes to ...

1. align the caret with the `match` keyword when pressing enter.

   The result is:
   ```
   - match: pattern
     |
   ```

   Before it was:

   ```
   - match: pattern
   |
   ```

   Note: This might also be achieved by indentation rules, but I had no luck with them so far.

2. trim spaces after `push: `, `set: `, ... if enter is pressed and adds an indented line afterwards.

   Note: The macro works only for this special purpose as `delete_word` also removes the colon.

3. removes automatic insertion of linefeed when typing colon after context names as it often ends in empty lines after editing context names.

### DIFF
--- a/Package/Default.sublime-keymap
+++ b/Package/Default.sublime-keymap
@@ -100,14 +100,6 @@
             { "key": "setting.packagedev.syntax.input_helpers" },
         ]
     },
-    {   // [optional] move caret to the next line and indent by one level.
-        "keys": ["enter"], "command": "insert", "args": {"characters": "\n  "}, "context": [
-            { "key": "preceding_text", "operator": "regex_match", "operand": "^ {4,}-\\s*match:.*$", "match_all": true },
-            { "key": "selection_empty", "match_all": true },
-            { "key": "selector", "operand": "source.yaml.sublime.syntax" },
-            { "key": "setting.packagedev.syntax.input_helpers" },
-        ]
-    },
     /* end of syntax_dev keys */
 
     /* start of syntax_test_dev keys */
--- a/Package/Default.sublime-keymap
+++ b/Package/Default.sublime-keymap
@@ -84,11 +84,27 @@
             { "key": "setting.packagedev.syntax.input_helpers" },
         ]
     },
-    {   // [optional] auto-insert newline after double-colon (contexts) and 'captures'
+    {   // [optional] auto-insert newline after double-colon of 'captures'
         "keys": [":"], "command": "insert", "args": { "characters": ":\n" }, "context": [
-            { "key": "preceding_text", "operator": "regex_match", "operand": "^ {0,2}[^\\s:]+$|^ {4,}(- *)?(captures)+$", "match_all": true },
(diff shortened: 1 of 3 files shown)


Package/Default.sublime-keymap
```diff
@@ -84,14 +84,30 @@
...
+        "keys": ["enter"],  "command": "run_macro_file", "args": {"file": "res://Packages/PackageDev/Package/Sublime Text Syntax Definition/Trim Add Newline.sublime-macro"}, "context": [
+            { "key": "preceding_text", "operator": "regex_match", "operand": "^ {4,}(?:- )?[^\\s:]+:\\s+$", "match_all": true },
+            { "key": "selection_empty", "match_all": true },
+            { "key": "selector", "operand": "source.yaml.sublime.syntax" },
+            { "key": "setting.packagedev.syntax.input_helpers" },
+        ]
+    },
+    {   // [optional] move caret to the next line and indent by one level.
+        "keys": ["enter"], "command": "insert", "args": {"characters": "\n  "}, "context": [
+            { "key": "preceding_text", "operator": "regex_match", "operand": "^ {4,}-\\s*match:.*$", "match_all": true },
```
FichteFoll, line 105: This definitely looks doable with indentation rules. I'll take a look later.

Package/Default.sublime-keymap
```diff
@@ -84,14 +84,30 @@
...
+    {   // [optional] auto-insert newline after double-colon of 'captures'
         "keys": [":"], "command": "insert", "args": { "characters": ":\n" }, "context": [
-            { "key": "preceding_text", "operator": "regex_match", "operand": "^ {0,2}[^\\s:]+$|^ {4,}(- *)?(captures)+$", "match_all": true },
+            { "key": "preceding_text", "operator": "regex_match", "operand": "^ {4,}(- *)?(captures)+$", "match_all": true },
             { "key": "selection_empty", "match_all": true },
             { "key": "selector", "operand": "meta.block.contexts.sublime-syntax" },
             { "key": "setting.packagedev.syntax.input_helpers" },
         ]
     },
+    {   // [optional] trim space after colon and insert newline.
```
FichteFoll, line 95: What would this be used for? In what situations? And why would trimming trailing whitespace by ST not be the better solution?

deathaxe: A space is added after colon when completing statements like `push:` or `set:`. That's fine if a context name or list is entered directly in the same line. But as soon as one presses enter to open up a new anonymous context or add a list of contexts to push/set, the space after the colon should be removed.

This may also be achieved via `trim_trailing_whitespace_on_save`, but as this setting is `false` by default, I ended up very often in leaving spaces after colons.

```
  - match: foo
    push: |

  - match: bar
    push:
      |
```

Package/Default.sublime-keymap
deathaxe: Finally found the Oniguruma Indentation Rules to be the one which is applied after match. So adding separate rules file didn't work.

Package/Default.sublime-keymap
FichteFoll: Thanks for the clarification.

PR #310, opened 2020-10-21
Color Scheme: Remove CSS.sublime-syntax#percentage-value

deathaxe:
It is already included in `CSS.sublime-syntax#numeric-values`.


FichteFoll:
I'll need to check if this was a more or less recent change and if we're not "breaking" older versions of the default CSS syntax with this. Tests will also need updating with an OR selector to cover both versions.

PR #316, opened 2020-11-20
Syntax: Fix reference to updated core YAML syntax

deathaxe:
Broken by: https://github.com/sublimehq/Packages/pull/2566


FichteFoll:
Thanks, I totally forgot that we used this here.

PR #318, opened 2020-12-06
Scopes: Fix completions

deathaxe:
Not only `sorted` is useless because of ST sorting completion suggestions by relevance it also breaks the function itself.

The scope names don't have kind info or annotation applied.

Removing `sorted` fixes that.


FichteFoll:
I tried getting ST to honor the provided sorting, but none of the options of `auto_complete_preserve_order` seemed to have an effect, so maybe there really is no point to sorting them.

PR #320, opened 2020-12-19
[ST4] Some changes to build system completions & updating it's syntax.

Ultra-Instinct-05:
This PR

1. Adds completions for `kill_previous` (introduced in 4092 to kill previously running builds) & `update_annotations_only` (introduced in 4050) to the build system completions file. These are also added to the syntax file for appropriate highlighting.
2. It removes any phantom related `exec` args from the completions & syntax file since phantoms are no longer used to show build errors in the buffer from 4050 onwards.


FichteFoll:
I'm wondering whether we should highlight these very specific args in build systems in the first place, since they are primarily used for other invocations of the `exec` command through key bindings or the command palette rather than a build system, but I'll decide on that later.

Thanks.

PR #330, opened 2021-01-31
[ST4] Make editing scheme/theme, OS aware when the corresponding is set to auto.

### LINKED ISSUE #329: Make Edit Theme & Scheme dark-/light-aware
https://github.com/sublimehq/sublime_text/issues/3884#issuecomment-769752092

> @FichteFoll You'll want to check out the new `sublime.ui_info()` API. Docs to come, but it returns a dict of info about the system style, and resolved theme and color scheme, along with the color palette from the color scheme.

I'll probably keep the quick panel prompt but set the selected index to the active one and add an "active" text either in the item itself or the annotation.

Ultra-Instinct-05:
This PR adds the ability to make editing color schemes/themes OS aware, when the corresponding settings are set to `auto`. The quick panel item that is active is selected and has an annotation of `Active` based on the OS appearance mode. 

Resolves #329

### DIFF
--- a/plugins/color_scheme_dev.py
+++ b/plugins/color_scheme_dev.py
@@ -278,10 +278,10 @@ class PackagedevEditSchemeCommand(sublime_plugin.WindowCommand):
                     self.open_scheme(paths[i][1])
 
             self.window.show_quick_panel(
-                choices, 
+                choices,
                 on_done,
-                selected_index = selected_index,
-                placeholder = 'Choose a color scheme to edit ...'
+                selected_index=selected_index,
+                placeholder='Choose a color scheme to edit ...'
             )
 
     @staticmethod
--- a/plugins/color_scheme_dev.py
+++ b/plugins/color_scheme_dev.py
@@ -256,16 +256,33 @@ class PackagedevEditSchemeCommand(sublime_plugin.WindowCommand):
                 (setting, self.get_scheme_path(view, setting))
                 for setting in ('dark_color_scheme', 'light_color_scheme')
             ]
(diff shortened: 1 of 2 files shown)


plugins/color_scheme_dev.py
```diff
@@ -256,16 +256,33 @@ def run(self):
...
+            for idx, choice in enumerate(choices):
+                if current_os_mode == 'dark' and choice.trigger == 'dark_color_scheme':
+                    choice.annotation = 'Active'
+                    selected_index = idx
+                elif current_os_mode == 'light' and choice.trigger == 'light_color_scheme':
+                    choice.annotation = 'Active'
+                    selected_index = idx
+                else:
+                    choice.annotation = ''
+                    selected_index = -1
```
FichteFoll, line 274: This `else` branch will result in the selected index being reset if the first case (dark) matches, after which the light case cannot match on the second iteration.

This would work better as follows:

```suggestion
            selected_index = -1
            for idx, choice in enumerate(choices):
                if choice.trigger.startswith(current_os_mode):
                    choice.annotation = 'Active'
                    selected_index = idx
```

FichteFoll (commented):
Thanks for the PR.

I'll hold off merging this until the next build is out and I can properly test, since I'm personally skipping 4096 due to the forced titlebars and I could see others doing that as well, meaning I can't only support the latest build.

plugins/color_scheme_dev.py
Ultra-Instinct-05: Thanks. Your solution is much cleaner. I have a lot to learn.

PR #332, opened 2021-01-31
Commands: Add support for the platform key

Ultra-Instinct-05:
This PR

1. Adds support for the `platform` key in `.sublime-commands` introduced in Build 4094 to support platform specific entries in the command palette.
2. Added completions for `platform` names.

This is essentially just a copy of what exists for the `.sublime-menu` file (Both the contexts and the completions). Maybe we need a way to reuse them ?

### DIFF
--- a/Package/Sublime Text Commands/Sublime Text Commands.sublime-syntax
+++ b/Package/Sublime Text Commands/Sublime Text Commands.sublime-syntax
@@ -93,7 +93,7 @@ contexts:
       scope: punctuation.definition.string.begin.json
       set:
         - meta_scope: meta.platform-name.sublime-commands string.quoted.double.json
-        - meta_content_scope: invalid.illegal.invalid-command.sublime-commands
+        - meta_content_scope: invalid.illegal.invalid-platform.sublime-commands
         - match: \"
           scope: punctuation.definition.string.end.json
           pop: true
--- a/Package/Sublime Text Commands/Sublime Text Commands.sublime-syntax
+++ b/Package/Sublime Text Commands/Sublime Text Commands.sublime-syntax
@@ -114,6 +114,12 @@ contexts:
   expect-colon:
     - include: Sublime JSON.sublime-syntax#expect-colon
 
+  expect-string-rest:
+    - include: Sublime JSON.sublime-syntax#expect-string-rest
+
+  string-escape:
+    - include: Sublime JSON.sublime-syntax#string-escape
+
   expect-string-value:
     - include: Sublime JSON.sublime-syntax#expect-string-value
 
--- a/Package/Sublime Text Commands/Sublime Text Commands.sublime-syntax
+++ b/Package/Sublime Text Commands/Sublime Text Commands.sublime-syntax
@@ -60,6 +60,13 @@ contexts:
         2: keyword.other.main.sublime-commands
         3: punctuation.definition.string.end.json
       set: [expect-mapping-value, expect-colon]
+    - match: (")(platform)(")
+      scope: meta.mapping.key.json meta.main-key.sublime-commands string.quoted.double.json
+      captures:
+        1: punctuation.definition.string.begin.json
+        2: keyword.other.main.sublime-commands
+        3: punctuation.definition.string.end.json
+      set: [expect-platform-string-value, expect-colon]
     - match: (")(\w*)(")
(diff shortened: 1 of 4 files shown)


Package/Sublime Text Commands/Sublime Text Commands.sublime-syntax
```diff
@@ -68,6 +75,33 @@ contexts:
...
+  platform-name-string-pop:
+    - match: (\")(!)?(Windows|Linux|OSX)(\")
+      scope: meta.platform-name.sublime-commands string.quoted.double.json
+      captures:
+        1: punctuation.definition.string.begin.json
+        2: keyword.operator.logical.sublime-commands
+        3: support.constant.command-name.sublime-commands
+        4: punctuation.definition.string.end.json
+      pop: true
+    - match: \" # probably illegal command name
```
FichteFoll, line 92: Incorrect comment.

Package/Sublime Text Commands/Sublime Text Commands.sublime-syntax
```diff
@@ -68,6 +75,33 @@ contexts:
...
+      scope: punctuation.definition.string.begin.json
+      set:
+        - meta_scope: meta.platform-name.sublime-commands string.quoted.double.json
+        - meta_content_scope: invalid.illegal.invalid-command.sublime-commands
```
FichteFoll, line 96: Incorrect scope.

FichteFoll (changes requested):
Thanks, just a few minor issues. Also, could you add a test, please?

Having platform matching twice is fine for me. While they are similar, I wouldn't say one is referencing the other and they are just using the same concept. We also wouldn't be able to use the correct scope suffix.

Package/Sublime Text Commands/Sublime Text Commands.sublime-syntax
Ultra-Instinct-05: This (& the below) is verbatim taken from the `.sublime-menu` syntax. I just changed the suffix scope to `.sublime-commands`. Does that mean it's wrong in `.sublime-menu` syntax ? (Perhaps I am mistaken)

Ultra-Instinct-05:
@FichteFoll regarding the tests, I wanted to add them but it seems the test file is a `.json` file. Does that mean I have to manually add the `.sublime-` specific scopes ?

Package/Sublime Text Commands/Sublime Text Commands.sublime-syntax
FichteFoll: Ah yes, it is probably wrong there as well, since this isn't about command names.

FichteFoll:
I don't understand the question regarding the test file? The file works like any other, but does not have the `.sublime-commands` extension so it isn't picked up by ST as a resource file.

Ultra-Instinct-05:
> it isn't picked up by ST as a resource file.

That makes sense, but what I am saying is suppose I have the `"platform"` in the `syntax_test_commands.json` and use PD to `^` and get the scope completions. It completes it as ` meta.sequence.json meta.mapping.key.json string.quoted.double.json` rather than `keyword.other.main.sublime-commands` (so the `.sublime-` specific scopes are missing). I was wondering if I am doing something wrong.

Package/Sublime Text Commands/Sublime Text Commands.sublime-syntax
Ultra-Instinct-05: Can you tell which is the incorrect scope. I can then add a commit that rectifies it for both syntaxes.

FichteFoll:
Did you set the file to use the correct syntax?

Package/Sublime Text Commands/Sublime Text Commands.sublime-syntax
Ultra-Instinct-05: This is resolved. Removed it from both menu and command syntax. Added tests as well.

Package/Sublime Text Commands/Sublime Text Commands.sublime-syntax
FichteFoll: `invalid-command` should be `invalid-platform`

FichteFoll:
I realized that the invalid scope probably makes the experience while typing quite uncomfortable and will likely tame that to a generic string instead.

PR #333, opened 2021-02-01
[ST4] Sync the in string completions to that of global keys completions for color scheme.

Ultra-Instinct-05:
This PR syncs the `Globals Keys (in-string).sublime-completions` to that of `Globals Keys.sublime-completions`.

Some of the missing keys were:
1. `block_caret_corner_style`
2. `block_caret_border`
3. `block_caret_underline`
4. `block_caret_corner_radius`

These were added in ac80cf7d9d681677b023d21d146143ef68f2c0c8 but looks like I forgot to add them to string completions.


FichteFoll:
Thanks. I thought I checked for those, but it seems I missed them as well.

PR #337, opened 2021-02-04
Remove the use of deprecated API endpoints.

### LINKED ISSUE #336: use of deprecated APIs
- use of `view.set_syntax_file()` should be replaced by `view.assign_syntax()`
- use of `sublime.status_message()` should be replaced by `window.status_message()`
- use of `window.get_output_panel()` should be replaced by `window.create_output_panel()` (compare `window.find_output_panel()`)

Ultra-Instinct-05:
This PR

1. Makes use of `Window.status_message` instead of `sublime.status_message`.
2. Uses `View.assign_syntax` instead of `View.set_syntax_file` (which uses `assign_syntax` internally).
3. I did not find any instances of usage of `Window.get_output_panel` apart from that in comments.

PS: I have also slipped in a commit to add completions for the `platform` key which was added in 3f858fd7997a8dd3ff85d5ac63584214983542ea (which I unfortunately forgot to add, but did not seem well worth a seperate PR).

Resolves #336

### DIFF
--- a/plugins/syntax_dev_legacy.py
+++ b/plugins/syntax_dev_legacy.py
@@ -25,9 +25,9 @@ __all__ = (
 PACKAGE_NAME = __package__.split('.')[0]
 
 
-def status(msg, console=False):
+def status(msg, window, console=False):
     msg = "[%s] %s" % (PACKAGE_NAME, msg)
-    sublime.status_message(msg)
+    window.status_message(msg)
     if console:
         print(msg)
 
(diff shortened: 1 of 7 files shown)


plugins/syntax_dev_legacy.py
```diff
@@ -255,7 +258,7 @@ def run(self, edit,
 
             if not text:
                 output.print("Error re-dumping the data in file (no output).")
-                status("Error re-dumping the data (no output).", True)
+                status("Error re-dumping the data (no output).", window, True)
```
FichteFoll, line 261: You updated all callsites of this function but didn't change the function itself.

FichteFoll (changes requested):

plugins/syntax_dev_legacy.py
Ultra-Instinct-05: I left out the `completions.py` as well. Hopefully, everything is resolved.

FichteFoll:
Thanks.

jrappen:
Thanks

PR #340, opened 2021-03-12
Trim redundant prefixes of syntax test suggestions.

Thom1729:
The implementation isn't necessarily final, but feedback would be appreciated.

First, a motivating example. The pipe denotes the cursor, and `syntax_test.suggest_scope_suffix` is false.

Before:

```js
    if ( true ) { }
//  ^^^^^^^^^^^^^^^ meta.conditional
//  ^^ keyword.control.conditional.if
//     |
```

Ordinary result of typing `^`:

```js
    if ( true ) { }
//  ^^^^^^^^^^^^^^^ meta.conditional
//  ^^ keyword.control.conditional.if
//     ^ meta.conditional meta.group punctuation.section.group.begin|
```

With `"syntax_test.suggest_trimmed_prefix": true`:

```js
    if ( true ) { }
//  ^^^^^^^^^^^^^^^ meta.conditional
//  ^^ keyword.control.conditional.if
//     ^ meta.group punctuation.section.group.begin|
```

---

The implementation is a first pass, but it's probably mostly correct. Notes:

- The setting name is off the top of my head and could be changed.
- It might be neater to add an `assertion_scopes` property to `AssertionLineDetails` rather than re-extracting that in the new code.
- The code assumes that assertions in previous lines are ordered. This should probably be changed.

### DIFF
--- a/Package/PackageDev.sublime-settings
+++ b/Package/PackageDev.sublime-settings
@@ -31,6 +31,10 @@
     // then suggest meta.example.markdown (true) vs meta.example (false).
     "syntax_test.suggest_scope_suffix": true,
 
+    // Whether to trim leading scopes from suggestions
+    // that are already part of previous assertions.
+    "syntax_test.suggest_trimmed_prefix": false,
+
 
     /// ADVANCED SETTINGS /////////////////////////////////////////////////////////////////////////
 
--- a/plugins/syntaxtest_dev.py
+++ b/plugins/syntaxtest_dev.py
@@ -352,6 +352,22 @@ class PackagedevSuggestSyntaxTestCommand(sublime_plugin.TextCommand):
         suggest_suffix = get_setting('syntax_test.suggest_scope_suffix', True)
         scope = find_common_scopes(scopes, not suggest_suffix)
 
+        trim_prefix = get_setting('syntax_test.suggest_trimmed_prefix', False)
+        if trim_prefix:
+            above_scopes = [
+                self.view.substr(sublime.Region(
+                    line.line_region.begin() + line.assertion_colrange[1],
+                    line.line_region.end(),
+                )).strip()
+                for line in lines[1:]
+                if line.assertion_colrange[0] <= lines[0].assertion_colrange[0]
+                and line.assertion_colrange[1] >= lines[0].assertion_colrange[1]
+            ]
+
+            for above_scope in reversed(above_scopes):
+                if scope.startswith(above_scope):
+                    scope = scope[len(above_scope):].strip()
+
         # delete the existing selection
         if not view.sel()[0].empty():
             view.erase(edit, view.sel()[0])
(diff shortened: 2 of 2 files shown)


deathaxe:
Asume your test cases are located within a `meta.block` because those rules would apply only there.

```
{
// <- meta.block

    if ( true ) { }
//  ^^^^^^^^^^^^^^^ meta.conditional
//  ^^ keyword.control.conditional.if
//     |
}
```

As `meta.block` is not of interest in test cases and therefore ommited, your algorithm currently fails in stripping `meta.condiational`. The result is as if the functionality wasn't present then.

```
{
    if ( true ) { }
//  ^^^^^^^^^^^^^^^ meta.conditional
//  ^^ keyword.control.conditional.if
//     ^ meta.conditional meta.group punctuation.section.group.begin|
}
```

It only works if `meta.block` is included in the assertion as follows.

```
{
// <- meta.block

    if ( true ) { }
//  ^^^^^^^^^^^^^^^ meta.block meta.conditional
//  ^^ keyword.control.conditional.if
//     |
}
```

The stripping algorithm should therefore lookup e.g. `meta.conditioanal` and remove all items to its left as well.

Thom1729:
It should handle that now. Some examples:

```js

{
    if ( true ) { }
//  ^^^^^^^^^^^^^^^ meta
//     ^ meta.block meta.conditional meta.group punctuation.section.group.begin
}

{
    if ( true ) { }
//  ^^^^^^^^^^^^^^^ meta.block
//     ^ meta.conditional meta.group punctuation.section.group.begin
}

{
    if ( true ) { }
//  ^^^^^^^^^^^^^^^ meta.block meta
//     ^ meta.conditional meta.group punctuation.section.group.begin
}

{
    if ( true ) { }
//  ^^^^^^^^^^^^^^^ meta.conditional
//     ^ meta.group punctuation.section.group.begin
}
```

Hopefully the implementation's not too abstruse.

deathaxe:
The algorithm seems to fail in case of more complicated test selectors such as

```
//                     ^^^^^^^^^^^^^^^^^^^^^^^ meta.function-call.arguments.scl - meta.path.plc
```

Thom1729:
Can you post the line of code being tested? Also, is that Verilog?

deathaxe:
It's SCL (Structured Control Language) a subset of ST (Structured Text) used to program PLCs. I am about to create syntax support for it. 

The code snippet being tested is:

```
    #struct.arrayFun[1](execute := #var + 10 );
//  ^^^^^^^ meta.path.plc variable.namespace.struct.plc
//         ^ meta.path.plc punctuation.accessor.dot.plc
//          ^^^^^^^^^^^ meta.path.plc meta.function-call.identifier.scl
//                     ^^^^^^^^^^^^^^^^^^^^^^^ meta.function-call.arguments.scl - meta.path
//                     ^ meta.block.fb.body.plc meta.path.plc meta.function-call.arguments.scl punctuation.section.group.begin.scl
```

Already got an idea of what's going wrong. Maybe nothing we can fix anyway. Due to an syntax issue the pattern `meta.function-call.arguments.scl - meta.path` fails to match. I wasn't aware of that detail on the time posting my comment.

FichteFoll:
Thanks for these changes. They seem to work pretty well for the 99% case and we can do follow-ups for edge cases later, if we need to. I also enabled this feature by default.

PR #341, opened 2021-03-13
Add command to autogenerate syntax tests for line of code.

Thom1729:
Example:

```js

    if ( true ) { }
```

Result (with `"syntax_test.suggest_scope_suffix": true`):

```js

    if ( true ) { }
//  ^^^^^^^^^^^^^^^ meta.conditional
//  ^^ keyword.control.conditional.if
//     ^^^^^^^^ meta.group
//     ^ punctuation.section.group.begin
//       ^^^^ constant.language.boolean.true
//            ^ punctuation.section.group.end
//              ^^^ meta.block
//              ^ punctuation.section.block.begin
//                ^ punctuation.section.block.end
```

Could probably use a key binding or something, but the implementation should be pretty much good.

### DIFF
--- a/plugins/syntaxtest_dev.py
+++ b/plugins/syntaxtest_dev.py
@@ -15,6 +15,7 @@ __all__ = (
     'PackagedevAlignSyntaxTestCommand',
     'PackagedevSuggestSyntaxTestCommand',
     'AssignSyntaxTestSyntaxListener',
+    'PackagedevGenerateSyntaxTestsForLineCommand',
 )
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


Thom1729:
I added some code to make some tests for nested scopes more compact.

Before:

```js
    42
//  ^^ meta.number.integer.decimal
//  ^^ constant.numeric.value
```

After:

```js
    42
//  ^^ meta.number.integer.decimal constant.numeric.value
```

FichteFoll:
That looks really nice. I'm going to give this a test run in hopefully the next week and work on another release.

Just wondering how we could make this command more accessible than just in the command palette without being obstrusive.

Thom1729:
I think it would make sense to bind it to control-enter or something. Do we have a context key for determining whether the view is a syntax test? If not, it would be easy to add.

FichteFoll:
The compacting code didn't work for me, but after some fiddling I was able to find the bugs and squash them. This will be included in the next release. I'm working on the changelog already. :rocket:

https://user-images.githubusercontent.com/931051/155862272-474a919a-1303-43aa-8e5c-af545f716bfe.mp4

PR #348, opened 2021-06-11
Settings: Fix completion values

### LINKED ISSUE #347: Wrong text alignment & cursor position after clicking phantom button in references editor
## Description

In preferences split editor, inline phantom button with pencil icon in the left pane sometimes misbehaves: when inserting new preferences entry to the right page, it may misalign text and cursor with selection. The bug seem to affect json map (object) values preferences. Note that there aren't any of such objects in the general _User/Preferences.sublime-settings_, so check out _Package Control.sublime-settings_ instead.

## Steps to reproduce

1. Open menu Preferences > Package Settings > Package Control -> Settings.
2. Split window should show up with default config on the left and user's prefs on the right.
3. Click phantom pencil icon on any preference which has JSON object default value on the left, and which is not yet defined on the right.

Before:
1-before

## Actual result

After:
2-after

## Expected result

Expected to see something similar to what happens with JSON arrays:
- new preference name on its own line, followed by double-colon,
- JSON object's curly braces on their own lines below,
- cursor between braces on its own line, indented.

Like this (cursor is denoted by `${|}`):
```
	"install_prereleases":
	[
		${|}
	],
```

## Version

Operating System: Arch Linux
Kernel Version: 5.12.9-arch1-1
OS Type: 64-bit
Graphics Platform: X11
Sublime Text: Stable, Build 4107

ratijas:
The root cause was that text manipulation was reordered in a wrong way:
JSON-encoded string was converted to snippet-field-content-encoded
string first, but then it was treated as a simply JSON-encoded string
again. Namely, stripping curly brackets away _after_ encoding them
as \\} broke cursor things in a weird ways.

Used pretty printing just because it looks better. But it required extra
whitespace stripping: both outside and inside stripped
brackets/braces.

Also changed variable name from `encoded` to `content` in fmt, so they
can never be accidentally reordered again.

Fixes #347

### DIFF
--- a/plugins/settings/known_settings.py
+++ b/plugins/settings/known_settings.py
@@ -434,6 +434,14 @@ class KnownSettings(object):
         return completions
 
     @staticmethod
+    def _encode_snippet_field_default_content(content: str) -> str:
+        """Escape string for snippet's default content context."""
+        return content             \
+            .replace("\\", "\\\\") \
+            .replace("$", "\\$")   \
+            .replace("}", "\\}")
+
+    @staticmethod
     def _key_snippet(key, value, bol="", eol=",\n"):
         """Create snippet with default value depending on type.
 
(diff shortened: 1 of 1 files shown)


ratijas:
ping @FichteFoll

FichteFoll:
Sorry for the delay. I took a very brief look but haven't found the time to finish up a release where this would then be included (I already added the milestone of the next version). I don't want to make promises, but I have some free time on Thursday where I hope to finally get to PD again.

ratijas:
take your time :)

FichteFoll:
Thanks, works really well and pretty-printing is a good idea, especially since we already insert the content in a new line.

PR #349, opened 2021-06-11
removed '#' from word_separators

MattDMo:
This will allow users to select an entire hex designation (`#AABBCC`) rather than just the alphanumeric part (`AABBCC`).


FichteFoll:
The setting currently matches what the default CSS settings do and I'd generally prefer to keep it that way.

PR #351, opened 2021-09-17
Fix syntax test tab dialog appearing on navigating test files using `Goto Anything`.

### LINKED ISSUE #177: You MUST use spaces! error: Syntax tests do not work properly with tabs as indentation.
So, I press `Ctrl+P` and start typing a file I would like to open, then BAM! You do not even let me complete type the file name `gotolasteditenhanced`.

image

I would happen to accidentally hit a syntax file which uses `tabs as indentation`:
```python
File: Data/Packages/PackageDev/plugins_/syntaxtest_dev.py
29: def _show_tab_warning():
30:     sublime.error_message("Syntax tests do not work properly with tabs as indentation."
31:                           "\n\nYou MUST use spaces!")
32: 
```
Your warning seem too aggressive popping up on a error message dialog. Also, very frustrating while using the `Ctrl+P`/Go To Anything... feature.

Why not just using `print` on the user console, perhaps also a `sublime.status_message` instead of `sublime.error_message` or `sublime.message_dialog`?

### LINKED ISSUE #350: PackageDev's Tab-Indentation-Error-Dialog interrupts Goto Anything
Each time a syntax test file being detected to use tab indentation is displayed in Goto Anything preview, the following error dialog is displayed and Goto Anything Quick Panel closes. 

That's quite annoying.

grafik

Can the check be disabled for transient views?

Ultra-Instinct-05:
This PR fixes a situation where navigating syntax test files using `Goto Anything` can be disruptive. Since the view is loaded and PD throws the dialog for those files that use tabs in test files, this results in a bad UX.

To prevent this, we simply skip files that are transiently loaded when navigating `Goto Anything`.

Fixes #350, fixes #177

### DIFF
--- a/plugins/syntaxtest_dev.py
+++ b/plugins/syntaxtest_dev.py
@@ -431,6 +431,14 @@ class AssignSyntaxTestSyntaxListener(sublime_plugin.EventListener):
         test_header = get_syntax_test_tokens(view)
         if not test_header:
             return
+
+        # If the user navigates through syntax test files using Goto Anything,
+        # then the navigation is disrupted (because of the dialog) if the test
+        # files happens to use indentation. So, we just skip checking any files
+        # that are transiently opened by using Goto Anything.
+        if view.sheet().is_transient():
+            return
+
         current_syntax = view.settings().get('syntax', None)
         test_syntax = test_header.syntax_file
 


keith-hall (approved):

FichteFoll:
I have two issues with this:

1. With where you inserted the check, automatic syntax assignment is skipped. I want to keep that.
2. We need a way to detect when the view has been properly opened and then run the check. I know it's annoying in general, but as I stated in #177 I *want* it to be annoying - just not while using Goto Anything and for transient views. As a general idea, I think we can start a `set_timeout` loop that checks whether a view id is still valid and if yes, whether the respective sheet isn't transient anymore.

PR #355, opened 2021-10-08
Handle syntax_test headers which use "partial-symbols" mode

keith-hall:
This updates the syntax test header line parsing to correctly recognize `partial-symbols` as valid, as per https://github.com/sublimehq/sublime_text/issues/3154#issuecomment-698774831

and also fixes a snippet syntax test which fails now that the Regular Expressions package doesn't scope `meta.literal` (see https://github.com/sublimehq/Packages/pull/3034/commits/0caecee8161c29e03d5bd3ac6ba87b41796f88c7)


FichteFoll:
Thanks, cherry-picked [`e32d061` (#355)](https://github.com/SublimeText/PackageDev/pull/355/commits/e32d061bfcd7e995a2c740a19e07f6fbebc7a311) since I handled the snippet regex thing in bf1b059b8f8ac872a8c79c5605749ccb456f05e7.

PR #357, opened 2021-10-30
Update color scheme word_separators, change LICENSE year

MattDMo:
I removed `#` from the color scheme `word_separators` so hex colors can be completely selected on double-click. I also updated the year in the `LICENSE` to 2021. This will need to be changed again in 2 months. Maybe somebody can write an action for that...

### DIFF
--- a/Package/Sublime Text Color Scheme/Sublime Text Color Scheme.sublime-settings
+++ b/Package/Sublime Text Color Scheme/Sublime Text Color Scheme.sublime-settings
@@ -5,5 +5,5 @@
         { "selector": "meta.mapping.key punctuation.definition.string.begin" },
     ],
     "extensions": ["sublime-color-scheme"],
-    "word_separators": "./\\()\"':,.;<>~!@$%^&*|+=[]{}`~?" // default without '-'
+    "word_separators": "./\\()\"':,.;<>~!@$%^&*|+=[]{}`~?" // default without '-' and '#'
 }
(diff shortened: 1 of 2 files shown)


Ultra-Instinct-05 (approved):

Package/Sublime Text Color Scheme/Sublime Text Color Scheme.sublime-settings
```diff
@@ -5,5 +5,5 @@
         { "selector": "meta.mapping.key punctuation.definition.string.begin" },
     ],
     "extensions": ["sublime-color-scheme"],
-    "word_separators": "./\\()\"':,.;<>~!@#$%^&*|+=[]{}`~?" // default without '-'
+    "word_separators": "./\\()\"':,.;<>~!@$%^&*|+=[]{}`~?" // default without '-'
```
michaelblyons, line 8: Update comment, too?

MattDMo: Just did

michaelblyons (approved):
LGTM.

I _do_ wish (now that I think about it) that you could have scope-sensitive word separators. Seems like a big ask of ST core, though.

MattDMo:
Yeah, I don't think we'll be seeing that anytime soon...

FichteFoll:
Thanks, let's hope this isn't too unexpected for users (since CSS doesn't do this either, at least at the moment).

PR #358, opened 2021-11-04
Fix syntax definition automatic assignment for syntax test files

keith-hall:
This fixes the bug whereby opening a `syntax_test_` file from `OverrideAudit: Open Resource` etc. would open without the syntax being assigned automatically from the syntax test header line.

The `on_load` event listener is fired before the file was read from the `.sublime-package` file, meaning the view is still empty, causing it to miss the syntax test header. So, this commit changes the behavior so that, if it sees that the file being loaded resides within the package catalog, and is empty, then it will re-try after a short delay.

### DIFF
--- a/plugins/syntaxtest_dev.py
+++ b/plugins/syntaxtest_dev.py
@@ -428,6 +428,13 @@ class AssignSyntaxTestSyntaxListener(sublime_plugin.EventListener):
     PLAIN_TEXT = "Packages/Text/Plain text.tmLanguage"
 
     def on_load(self, view):
+        if view.size() == 0 and view.file_name().startswith(sublime.packages_path() + '/'):
+            sublime.set_timeout(lambda: self.assign_syntax(view), 100)
+            return
+
+        self.assign_syntax(view)
+
+    def assign_syntax(self, view):
         test_header = get_syntax_test_tokens(view)
         if not test_header:
             return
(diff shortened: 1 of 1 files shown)


plugins/syntaxtest_dev.py
```diff
@@ -428,6 +428,13 @@ class AssignSyntaxTestSyntaxListener(sublime_plugin.EventListener):
     PLAIN_TEXT = "Packages/Text/Plain text.tmLanguage"
 
     def on_load(self, view):
+        if view.size() == 0 and view.file_name().startswith(sublime.packages_path() + '/'):
+            sublime.set_timeout(lambda: self.assign_syntax(view), 100)
+            return
+
+        self.assign_syntax(view)
```
michaelblyons, line 435: Add an `else` here?

keith-hall: done :)

plugins/syntaxtest_dev.py
```diff
@@ -428,6 +428,12 @@ class AssignSyntaxTestSyntaxListener(sublime_plugin.EventListener):
     PLAIN_TEXT = "Packages/Text/Plain text.tmLanguage"
 
     def on_load(self, view):
+        if view.size() == 0 and view.file_name().startswith(sublime.packages_path() + '/'):
+            sublime.set_timeout(lambda: self.assign_syntax(view), 100)
```
FichteFoll, line 432: Hard timeouts like this are usually a thing we should try to avoid. Could we instead loop the function to itself and check for `view.is_loading()` instead of `view.size() == 0`? Unfortunately I wasn't able to reliably reproduce the issue to check whether that also works myself.

```suggestion
        if view.is_loading() and view.file_name().startswith(sublime.packages_path() + '/'):
            sublime.set_timeout(lambda: self.on_load(view), 100)
```

FichteFoll: Was able to reproduce. Unfortuantely, `view.is_loading` is `False` despite the view still being empty, so my idea doesn't work here.
I honestly wish `View.is_loading` would be useful *anywhere*, but so far all I can remember is getting disappointed by it.

PR #361, opened 2022-01-21
Added `disabled` theme attribute

AmjadHD:
Introduced in ST 4121, and not yet documented sublimehq/sublime_text/issues/5116.


FichteFoll:
Thanks!

PR #367, opened 2022-03-04
Don't restrict plugin reloading to Package Control Updates

deathaxe:
This commit enables reloading all sub modules by just saving `main.py`.

There's no real need to restrict plugin reloading to Package Control
updates. Reloading should work properly without Package Control as well.

That strategy is already used successfully by several packages, such as
A File Icon, GitGutter, MarkdownEditing and maybe more.

### DIFF
--- a/main.py
+++ b/main.py
@@ -1,21 +1,13 @@
-try:
-    from package_control import events
-except ImportError:
-    pass
-else:
-    if events.post_upgrade(__package__):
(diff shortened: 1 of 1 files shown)


FichteFoll:
A bit weird to unset `prefix` afterwards, but thanks.

deathaxe:
It enables garbage collecting the no longer needed global.

FichteFoll:
`del prefix` would have been preferred in that case.

PR #372, opened 2022-03-27
Syntax: Add syntax based folding support

deathaxe:
This PR adds support for the following tmPreferences keys

- indentationFoldingEnabled
- scopeFoldingEnabled
- foldScopes

Some effort has been made to avoid highlighting the whole file invalid in various incomplete code scenarios.

### DIFF
--- a/Package/TextMate Preferences/Completions/Settings Keys.sublime-completions
+++ b/Package/TextMate Preferences/Completions/Settings Keys.sublime-completions
@@ -78,6 +78,21 @@
 			"kind": ["snippet", "s", "setting"],
 		},
 		{
+			"trigger": "indentationFoldingEnabled",
+			"contents": "<key>indentationFoldingEnabled</key>\n<false/>",
+			"kind": ["snippet", "s", "setting"],
+		},
+		{
+			"trigger": "scopeFoldingEnabled",
+			"contents": "<key>scopeFoldingEnabled</key>\n<false/>",
+			"kind": ["snippet", "s", "setting"],
+		},
+		{
+			"trigger": "foldScopes",
+			"contents": "<key>foldScopes</key>\n<array>\n\t$0\n</array>",
+			"kind": ["snippet", "s", "setting"],
+		},
+		{
 			"trigger": "icon",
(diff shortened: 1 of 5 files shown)


Package/TextMate Preferences/Completions/Settings Keys.sublime-completions
```diff
@@ -77,6 +77,21 @@
 			"contents": "<key>preserveIndent</key>\n<string>$1</string>",
 			"kind": ["snippet", "s", "setting"],
 		},
+		{
+			"trigger": "indentationFoldingEnabled",
+			"contents": "<key>indentationFoldingEnabled</key>\n<false/>",
```
Ultra-Instinct-05, line 82: As per what was told in Discord, `indentationFoldingEnabled` is `<true />` by default. Hence, I guess it doesn't make sense for the snippet to put a `<false />` as completion.

Package/TextMate Preferences/Completions/Settings Keys.sublime-completions
```diff
@@ -77,6 +77,21 @@
...
+			"kind": ["snippet", "s", "setting"],
+		},
+		{
+			"trigger": "scopeFoldingEnabled",
+			"contents": "<key>scopeFoldingEnabled</key>\n<false/>",
```
Ultra-Instinct-05, line 87: Same as previous.

Package/TextMate Preferences/Completions/Settings Keys.sublime-completions
Ultra-Instinct-05: Reason why I feel `true` makes more sense is:
- It's the default value. So having it set to `<true />` or not having the entire `<indentationFoldingEnabled><true />` is the same and hence causes the least surprise.
- The user may not know that `<true />` is the default.

jrappen: Make it `... <${1:true}/ >`, which should make (1) it clear that true is the default and (2) editing easy.

deathaxe:
A full rework of Property Lists and tmPreferences is in progress.

PR #373, opened 2022-03-27
Syntax: Tweak shellVariables patterns in tmPreferences

deathaxe:
This commit removes some unnecessary push-pop operations from tmPreferneces files.


FichteFoll:
Please address the test failures. I confirmed that they were caused by this commit (as they don't fail without the commit).

deathaxe:
A full rework of Property Lists and tmPreferences is in progress.

PR #375, opened 2022-05-13
Syntax: Fix missing meta.string.yaml scope

jfcherng:


keith-hall (approved):

FichteFoll:
That was actually done on purpose back then because it would colorize, well, everything as string and would reduce the color spectrum for the embedded regular expressions. For consistency, the same would have to be applied to the unquoted strings, too.
Perhaps a `meta.string` scope would be an alternative?

jfcherng:
Interesting. So I guess it's designed that double quotes are highlighted as plain string, correct?

FichteFoll:
> Interesting. So I guess it's designed that double quotes are highlighted as plain string, correct?

It's more of me being lazy and not wanting to do yet another version of doubly-escaped regular expression shenanigans as with JSON-tmLanguage before. Double-quoted strings in YAML bring absolutely no advantage for syntax definitions outside of very small and almost plain-text matches beginning with a `'` character (so matching single-quoted strings, basically), in which case rich regex matching is just not needed.

PR #389, opened 2023-05-12
Add workaround for `scope_name()` bug

deathaxe:
see: https://github.com/sublimehq/sublime_text/issues/5971

In order to get correct suggestions, a small delay is required for ST to complete lexing. Otherwise `scope_name()` returns the wrong scopes.


keith-hall (approved):

FichteFoll:
I'm sorry that I haven't been able to look into this earlier.

I tested this workaround locally (using the steps for reproduction provided in the references ST bug report) and it seems to be rather unreliable, only occasionally picking up the correct scopes. Since this appears to be a timing issue and I don't want to delay the completion by an arbitrary amount (in addition to creating two undo steps instead of one), I'm opting to not include this workaround. Ideally it would be fixed in core …

I assume this is also why the PR was closed, but a short comment about it would have been appreciated.

PR #397, opened 2024-07-21
[YAML-tmLanguage] Remove empty capture group from multi-line match rule

radium-v:
Fixes the rule for matching `match`, `begin`, and `end` multi-line blocks.


keith-hall:
Thanks for your contribution. I'd be more confident about approving this if we had some syntax tests for YAML-tmLanguage... I will try to find time to try this change out locally within the next few days. Someone may beat me to it and that's fine 😉

FichteFoll (approved):
Thanks for taking a look at this very old comment that I totally do not remember writing but totally did (c6a170a42a26da4c30195ad025cead0ad0a1d190).

I haven't looked at this syntax in years and only verified that your change doesn't break anything obvious, which it did not.

Issue #398, opened 2024-10-20
Auto generated syntax tests shouldn't fill the gaps


deathaxe:
The command `packagedev_generate_syntax_tests_for_line` creates a single assertion for `meta.tag` in following example, while it should create two separate ones for opening and closing tags.

```
    <p>Our website is in maintenance mode. Please, come back later.</p>
{#^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.template.twig #}
{#  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.tag.block.any.html #}
{#  ^ punctuation.definition.tag.begin.html #}
{#   ^ entity.name.tag.block.any.html #}
{#    ^ punctuation.definition.tag.end.html #}
{#                                                                 ^^ punctuation.definition.tag.begin.html #}
{#                                                                   ^ entity.name.tag.block.any.html #}
{#                                                                    ^ punctuation.definition.tag.end.html #}
```

That always happens if tokens of same scope appear on a line. Created tests expect the whole region between start of first and end of last token to be scoped. Gaps are ignored.

PR #404, opened 2025-02-10
Fix auto-generated syntaxes not filling gaps (#398)

camila314:
Old Version (incorrectly assumes the space has numerical scopes):
```cpp
    (123 456)
/*  ^^^^^^^^^ meta.group.c-2 */
/*  ^ punctuation.section.group.begin */
/*   ^^^^^^^ meta.number.integer.decimal.c-2 constant.numeric.value.c-2 */
/*          ^ punctuation.section.group.end */```
```
New Version (correctly excludes space from numerical scopes):
```cpp
    (123 456)
/*  ^^^^^^^^^ meta.group.c-2 */
/*  ^ punctuation.section.group.begin */
/*   ^^^ meta.number.integer.decimal.c-2 constant.numeric.value.c-2 */
/*       ^^^ meta.number.integer.decimal.c-2 constant.numeric.value.c-2 */
/*          ^ punctuation.section.group.end */
```

Simply checks if the previous region's end matches with the new region's beginning when inserting into a ScopeTreeNode.

### DIFF
--- a/plugins/syntaxtest_dev.py
+++ b/plugins/syntaxtest_dev.py
@@ -535,6 +535,7 @@ class ScopeTreeNode:
 
     @classmethod
     def build_forest(cls, tokens, *, trim_suffix=False):
+        print(tokens)
         tokens = [
             (region, cls._split_scope(scope, trim_suffix=trim_suffix))
             for region, scope in tokens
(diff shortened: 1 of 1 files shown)


plugins/syntaxtest_dev.py
```diff
@@ -535,6 +535,7 @@ def __init__(self, region, scope, children=None):
 
     @classmethod
     def build_forest(cls, tokens, *, trim_suffix=False):
+        print(tokens)
```
michaelblyons, line 538: Did you mean to take this out?

camila314: no

michaelblyons:
I also recommend adding "Fixes #398” to the OP, and changing the title to something like "Fix automatic syntax assertions filling gaps" or "Stop automatic syntax tests missing gaps".

michaelblyons (approved):

FichteFoll (approved):
Thanks for taking a look at this issue and fixing it.

PR #426, opened 2026-04-05
Fix python 3.14 compatibility of settings helpers

deathaxe:
This commit replaces a deprecated method call, which is removed by python 3.14.


keith-hall (approved):
I checked to see if this would be backwards compatible with Python 3.8, and saw it was introduced in Python 3.4 👍
https://docs.python.org/3.8/library/html.html

PR #428, opened 2026-04-08
Add child combinator operator

deathaxe:
This commit adds...

1. a common `.combinator` 3rd-level scope, following the schema of CSS.
2. the `>` child combinator operator supported as of ST4201.

### DIFF
--- a/Package/Scope Selector.sublime-syntax
+++ b/Package/Scope Selector.sublime-syntax
@@ -45,15 +45,17 @@ contexts:
     - match: \s+\(
       scope: invalid.illegal.operator-required-between-scope-and-group.scope-selector
     - match: \s+(?=\w)
-      scope: keyword.operator.right.scope-selector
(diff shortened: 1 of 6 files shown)


Package/Scope Selector.sublime-syntax
```diff
@@ -45,15 +45,17 @@ contexts:
     - match: \s+\(
       scope: invalid.illegal.operator-required-between-scope-and-group.scope-selector
     - match: \s+(?=\w)
-      scope: keyword.operator.right.scope-selector
+      scope: keyword.operator.combinator.right.scope-selector
```
michaelblyons: If we're changing all of them, what do you think about using

```suggestion
      scope: keyword.operator.combinator.descendent.scope-selector
```

to match the `.child` for `>`?

FichteFoll (approved):

PR #430, opened 2026-04-08
Use safe_load in commandinfo

EhteshamSid:
`yaml.load()` in `plugins/command_completions/commandinfo.py` doesn't pass a safe Loader. This can deserialize arbitrary Python objects and is an RCE risk if the YAML comes from user input or the network. Switched to `yaml.safe_load()`.


EhteshamSid:
Closing - the patch introduced a typo (`yaml.safe_safe_load_all` is not a real function). Also, `SaveOmapLoader` inherits from `SafeConstructor` so the original code was already safe. Apologies for the noise.

PR #433, opened 2026-05-14
Adjust some scopes in Sublime Syntax Definitions

deathaxe:
This PR proposes to...

1. introduce unique scopes for context, branch_point and global variables.

    - `variable.other.branch_point` - branch point references
    - `variable.other.constant` - global variables
    - `variable.other.context` - context names
    
   Goal: Plugins, snippets and completions should be able to address all of them easily.
 
2. scopes `{{...}}` expressions `meta.interpolation` instead of `keyword.other.variable`.
  
3. change scope of context name from `entity.name.function` to `entity.name.class`

   to ensure _Goto Symbol_ quick panel displays `c` kind icon, as auto-completion popup does.
   
   The completion items' kind icon is changed to use KIND_ID_TYPE, to ensure both, 
   _Goto Symbol_ quick panel and auto-completion popup display `c` with same color.
   
   Goal: Better visual consistency.

### DIFF
--- a/Package/Sublime Text Syntax Definition/References - Variable.tmPreferences
+++ b/Package/Sublime Text Syntax Definition/References - Variable.tmPreferences
@@ -3,7 +3,7 @@
 <plist version="1.0">
 <dict>
     <key>scope</key>
-    <string>source.yaml.sublime.syntax keyword.other.variable variable.other</string>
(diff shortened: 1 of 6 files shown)


FichteFoll (approved):
Good changes. The variable scopes have been on my list for a long time but I never pushed for it.

PR #442, opened 2026-07-11
Fix invalid reference in file conversion module

### LINKED ISSUE #441: File conversion regression in v3.7.0
Following traceback is printed when calling `PackageDev: Convert (YAML, JSON, PList) to...` from command palette.

```
Traceback (most recent call last):
  File "C:\Apps\Sublime Text\Lib\python314\sublime_plugin.py", line 1596, in run_
    return self.run()
           ~~~~~~~~^^
  File "C:\Apps\Sublime Text\Data\Packages\PackageDev\plugins\file_conversion.py", line 188, in run
    for itm in self.target_list:
               ^^^^^^^^^^^^^^^^
AttributeError: 'PackagedevConvertCommand' object has no attribute 'target_list'
```

deathaxe:
Fixes #441

### DIFF
--- a/plugins/file_conversion.py
+++ b/plugins/file_conversion.py
@@ -185,7 +185,7 @@ class PackagedevConvertCommand(sublime_plugin.WindowCommand):
                     # Show overlay with all dumping options except for the current type
                     # Save stripped-down `items` for later
                     options, items = [], []
-                    for itm in self.target_list:
+                    for itm in target_list:
                         # To not clash with function-local "target_format"
                         target_format_ = itm['kwargs']['target_format']
                         if target_format_ != source_format:


FichteFoll:
I should really get some linting going on to prevent these obvious errors.

FichteFoll:
Hm, to spot this error I need to have sublime type stubs because this is a subclass of `sublime_plugin.WindowCommand`. Alas, type checking will be added in a PR soon.

deathaxe:
Well, LSP-basedpyright highlighted it pretty reliably, with type hints available via default API.

FichteFoll:
Yeah, it's those type hints I did not have because I was trying out `ty`.

